PR #29, opened 2024-03-22
add logger class

Bechir-Brahem:
Logging class that can log on different levels of verbosity (debug, info, warn, error)

log function should act like python's print function and takes any number/type of parameters

currently missing tests , ~~writing in files~~ , ~~printing from where the logging was called~~


erikfrojdh (approved):
Looks good. I'm not sure we should use it everywhere but for debugging during development it could be useful!

PR #39, opened 2024-04-02
moving code into aare:: namespace

erikfrojdh:
I think it's better to do this now than later when we have more code. Moved
- core
- file_io
into the aare namespace


Bechir-Brahem (approved):
LGTM

PR #40, opened 2024-04-02
add github ci

Bechir-Brahem:
build and test project with github actions
- use mamba when we use system libraries
- cache mamba envs
squash of #38


erikfrojdh (changes requested):
Could you disable python bindings for now? It would be nice if the checks are passing.

Bechir-Brahem:
> Could you disable python bindings for now? It would be nice if the checks are passing.

done

erikfrojdh (approved):

Issue #51, opened 2024-04-03
Read cluster files


erikfrojdh:
Not sure we can conform to the FileInterface API since we don't have strictly frame based data. 

Current implementation to merge with aare: https://git.psi.ch/sls_detectors_software/python_cluster_reader

PR #60, opened 2024-04-16
read write cluster file

Bechir-Brahem:
- related to #51 
- read and write simple cluster files


file_io/include/aare/file_io/ClusterFile.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+  private:
+    FILE *fp = nullptr;
+    // size_t current_cluster{};
+    std::filesystem::path fname{};
+    std::string mode{};
+    int32_t frame_number{};
+    int32_t n_clusters{};
+    static const int HEADER_BYTES = 8;
+};
```
Bechir-Brahem, line 68: maybe we can use FileInterface and make it a template to the type of object to operate on (Frame,Cluster...)?

Bechir-Brahem: `class ClusterFile: FileInterface<Cluster>`

PR #75, opened 2024-06-20
Feature/compile on windows

Bechir-Brahem:


Bechir-Brahem:
see https://github.com/slsdetectorgroup/aare/actions/runs/9598271744/job/26469161775?pr=75 for what should be fixed to make it compile on windows

PR #105, opened 2024-11-26
read_n returns remaining frames

erikfrojdh:
Modified read_n to return the number of frames available if less than the number of frames requested. 

```python
#f is a CtbRawFile containing 10 frames

f.read_n(7) # you get 7 frames
f.read_n(7) # you get 3 frames
f.read_n(7) # RuntimeError
```

Also added support for chunk_size when iterating over a file:

```python
# The file contains 10 frames


with CtbRawFile(fname, chunk_size = 7) as f:
    for headers, frames in f:
        #do something with the data
        # 1 iteration 7 frames
        # 2 iteration 3 frames
        # 3 iteration stops
```

### DIFF
--- a/python/aare/CtbRawFile.py
+++ b/python/aare/CtbRawFile.py
@@ -2,6 +2,7 @@
 from . import _aare
 import numpy as np
 from .ScanParameters import ScanParameters
+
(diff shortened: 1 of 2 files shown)


JulianHeymes (approved):
Working as expected

PR #137, opened 2025-03-17
Interpolate

erikfrojdh:
- added interpolation


src/Interpolator.cpp
```diff
@@ -0,0 +1,145 @@
...
+                           NDView<double, 1> ybins, NDView<double, 1> ebins)
+    : m_ietax(etacube), m_ietay(etacube), m_etabinsx(xbins), m_etabinsy(ybins), m_energy_bins(ebins) {
+    if (etacube.shape(0) != xbins.size() || etacube.shape(1) != ybins.size() ||
+        etacube.shape(2) != ebins.size()) {
+        throw std::invalid_argument(
+            "The shape of the etacube does not match the shape of the bins");
+    }
+
+    // Cumulative sum in the x direction, can maybe be combined with a copy?
+    for (ssize_t k = 0; k < m_ietax.shape(2); k++) {
```
AliceMazzoleni99, line 16: Reorder loop? According to c_strides first dimension is not contiguous in memory - maybe im wrong

include/aare/ClusterVector.hpp
```diff
@@ -231,6 +231,10 @@ template <typename T, typename CoordType = int16_t> class ClusterVector {
         return *reinterpret_cast<V *>(element_ptr(i));
     }
 
+    template <typename V> const V &at(size_t i) const {
```
AliceMazzoleni99, line 234: I would not template this function. It can be accessed from outside the class but the type V is specific to the class object e.g. if cluster_size_x is 3 or 2 or n. I would use a switch statement and return Cluster3x3 or Cluster2x2. Not that elegant to generalize but safer to use. Maybe an error or warning would be reasonable as well.

src/ClusterFile.cpp
```diff
@@ -333,6 +418,19 @@ Eta2 calculate_eta2(Cluster3x3 &cl) {
     return eta;
 }
 
+
+Eta2 calculate_eta2(Cluster2x2 &cl) {
+    Eta2 eta{};
+
```
AliceMazzoleni99, line 424: Maybe check if cl.data[0] + cl.data[1] is zero as f0r Cluster3x3

include/aare/algorithm.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include <vector>
+#include <aare/NDArray.hpp>
+
+namespace aare {
+/**
+ * @brief Find the index of the last element smaller than val
+ * assume a sorted array
+ */
+template <typename T>
+size_t last_smaller(const T* first, const T* last, T val) {
```
AliceMazzoleni99, line 14: For me not quite clear from function interpolation. Does not need to be strictly smaller, right?

src/algorithm.test.cpp
```diff
@@ -0,0 +1,63 @@
...
+    }
+    // arr 0, 1, 2, 3, 4
+    REQUIRE(aare::nearest_index(arr, 2.3) == 2);
+    REQUIRE(aare::nearest_index(arr, 2.6) == 3);
+    REQUIRE(aare::nearest_index(arr, 45.0) == 4);
+    REQUIRE(aare::nearest_index(arr, 0.0) == 0);
+    REQUIRE(aare::nearest_index(arr, -1.0) == 0);
+}
+
+TEST_CASE("Passing integers to nearest_index works"){
```
AliceMazzoleni99, line 20: Maybe add the same tag [algorithm] for all tests

src/Interpolator.cpp
```diff
@@ -0,0 +1,145 @@
...
+        for (ssize_t j = 0; j < m_ietax.shape(1); j++) {
+            for (ssize_t i = 1; i < m_ietax.shape(0); i++) {
+                m_ietax(i, j, k) += m_ietax(i - 1, j, k);
+            }
+        }
+    }
+
+    // Normalize by the highest row, if norm less than 1 don't do anything
```
AliceMazzoleni99, line 24: Instead of looping twice, first calculate last row of m_iatex and normalize directly when calculating m_iatex

src/Interpolator.cpp
```diff
@@ -0,0 +1,145 @@
...
+            // auto ie = nearest_index(m_energy_bins, photon.energy)-1;
+            // auto ix = nearest_index(m_etabinsx, eta.x)-1;
+            // auto iy = nearest_index(m_etabinsy, eta.y)-1;  
+            //Finding the index of the last element that is smaller
+            //should work fine as long as we have many bins
+            auto ie = last_smaller(m_energy_bins, photon.energy);
+            auto ix = last_smaller(m_etabinsx, eta.x);
+            auto iy = last_smaller(m_etabinsy, eta.y); 
+
+            photon.x += m_ietax(ix, iy, 0)*2; //eta goes between 0 and 1 but we could move the hit anywhere in the 2x2
```
AliceMazzoleni99, line 132: 0 should be ie

src/Interpolator.cpp
```diff
@@ -0,0 +1,145 @@
...
+            case cBottomLeft:
+                dX = -1.;
+                dY = -1.;
+                break;
+            case cBottomRight:
+                dX = 0.; 
+                dY = -1.;
+                break;
+            }
+            photon.x += m_ietax(ix, iy, 0)*2 + dX;
```
AliceMazzoleni99, line 107: 0 should be ie

src/Interpolator.cpp
```diff
@@ -0,0 +1,145 @@
...
+    photons.reserve(clusters.size());
+
+    if (clusters.cluster_size_x() == 3 || clusters.cluster_size_y() == 3) {
+        for (size_t i = 0; i<clusters.size(); i++){
+
+            auto cluster = clusters.at<Cluster3x3>(i);
+            Eta2 eta= calculate_eta2(cluster);
+    
+            Photon photon;
+            photon.x = cluster.x;
```
AliceMazzoleni99, line 68: probably misunderstanding, but why is the photon.x a double thought it should be an index. Same for photon.y

src/ClusterFile.cpp
```diff
@@ -108,6 +108,79 @@ ClusterVector<int32_t> ClusterFile::read_clusters(size_t n_clusters) {
...
 }
 
+ClusterVector<int32_t> ClusterFile::read_clusters(size_t n_clusters, ROI roi) {
+    if (m_mode != "r") {
+        throw std::runtime_error("File not opened for reading");
+    }
+    
+    ClusterVector<int32_t> clusters(3,3);
+    clusters.reserve(n_clusters);
+
```
AliceMazzoleni99, line 118: Dont understand what m_num_left, nn and nph should be

src/Interpolator.cpp
```diff
@@ -0,0 +1,145 @@
...
+            int ex, ey;
+            // cBottomLeft = 0,
+            // cBottomRight = 1,
+            // cTopLeft = 2,
+            // cTopRight = 3
+            switch (eta.c) {
+            case cTopLeft:
+                dX = -1.;
+                dY = 0;
+                break;
```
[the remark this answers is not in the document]
erikfrojdh: Agree, fixed

src/Interpolator.cpp
erikfrojdh: Good catch! Fixed!

src/Interpolator.cpp
erikfrojdh: Fixed

src/Interpolator.cpp
```diff
@@ -0,0 +1,145 @@
...
+                dY = -1.;
+                break;
+            case cBottomRight:
+                dX = 0.; 
+                dY = -1.;
+                break;
+            }
+            photon.x += m_ietax(ix, iy, 0)*2 + dX;
+            photon.y += m_ietay(ix, iy, 0)*2 + dY;
+            photons.push_back(photon);
```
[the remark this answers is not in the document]
erikfrojdh: Yes, the photon energy we can measure directly so this we don't need to interpolate. It's just the sum of the pixel values.

src/Interpolator.cpp
erikfrojdh: For the initial cluster we give the position (x,y) as pixel index and there it is an integer. But after interpolation we know the position with subpixel resolution. Meaning a cluster at (10,7) we can interpolate to (10.32, 7.63) in coordinates of the original image. 

Then at a later step we create the new image and then assign the hit to a pixel.

src/Interpolator.cpp
erikfrojdh: I'm not sure this will be faster but happy to benchmark at some later point.

My arguments are:
- You need to touch all pixels to calculate the last row 
- Since we are doing the cumulative sum we would either need out of order access or keep a temporary variable per column to avoid extra divides which are costly

src/ClusterFile.cpp
erikfrojdh: Fixed

src/algorithm.test.cpp
erikfrojdh: Fixed

include/aare/algorithm.hpp
erikfrojdh: Added test.

AliceMazzoleni99 (approved):

PR #143, opened 2025-03-25
added roi, noise and gain

### LINKED ISSUE #90: Cluster file reading and writing

### LINKED ISSUE #135: Configuration fails with cmake <3.28
EXCLUDE_FROM_ALL 1 causes cmake not to find target lmfit 

Works with newer cmake 

Options: find a fix or bump cmake requirement to 3.28

### LINKED ISSUE #139: Gain correction while reading cluster files
- apply a gain map when reading from a cluster file

erikfrojdh:
- Moved definitions of Cluster_2x2 and Cluster_3x3 to it's own file
- Added optional members for ROI, noise_map and gain_map in ClusterFile

**API:**

After creating the ClusterFile the user can set one or all of: roi, noise_map, gain_map

```python
f = ClusterFile(fname)
f.set_roi(roi) #aare.ROI
f.set_noise_map(noise_map) #numpy array
f.set_gain_map(gain_map) #numpy array
```

**When reading clusters they are evaluated in the order:**

1. If ROI is enabled check that the cluster is within the ROI
1. If noise_map is enabled check that the cluster meets one of the conditions 
    - Center pixel above noise
    - Highest 2x2 sum above 2x noise
    - 3x3 sum above 3x noise
  1. If gain_map is set apply the gain map before returning the clusters (not used for noise cut)

**Open questions:**
1. Check for out of bounds access in noise and gain map?

closes #139 
closes #135 
closes #90

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -396,6 +396,7 @@ if(AARE_TESTS)
             ${CMAKE_CURRENT_SOURCE_DIR}/src/NDView.test.cpp
             ${CMAKE_CURRENT_SOURCE_DIR}/src/ClusterFinder.test.cpp
             ${CMAKE_CURRENT_SOURCE_DIR}/src/ClusterVector.test.cpp
+            ${CMAKE_CURRENT_SOURCE_DIR}/src/ClusterFile.test.cpp
             ${CMAKE_CURRENT_SOURCE_DIR}/src/Pedestal.test.cpp
             ${CMAKE_CURRENT_SOURCE_DIR}/src/NumpyFile.test.cpp
             ${CMAKE_CURRENT_SOURCE_DIR}/src/NumpyHelpers.test.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,12 +81,12 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 if(AARE_FETCH_LMFIT)
     #TODO! Should we fetch lmfit from the web or inlcude a tar.gz in the repo?
-    set(lmfit_patch git apply ${CMAKE_CURRENT_SOURCE_DIR}/patches/lmfit.patch)
+    set(LMFIT_PATCH_COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/patches/lmfit.patch)
     FetchContent_Declare(
         lmfit
         GIT_REPOSITORY https://jugit.fz-juelich.de/mlz/lmfit.git
         GIT_TAG        main
-        PATCH_COMMAND ${lmfit_patch}
+        PATCH_COMMAND ${LMFIT_PATCH_COMMAND}
         UPDATE_DISCONNECTED 1
         EXCLUDE_FROM_ALL 1
     )
@@ -111,10 +111,12 @@ if(AARE_FETCH_ZMQ)
     if (${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.30")
         cmake_policy(SET CMP0169 OLD)
     endif()
+    set(ZMQ_PATCH_COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/patches/libzmq_cmake_version.patch)
     FetchContent_Declare(
         libzmq
         GIT_REPOSITORY https://github.com/zeromq/libzmq.git
         GIT_TAG        v4.3.4
+        PATCH_COMMAND ${ZMQ_PATCH_COMMAND}
     )
     # Disable unwanted options from libzmq
     set(BUILD_TESTS OFF CACHE BOOL "Switch off libzmq test build")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,14 +82,28 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 if(AARE_FETCH_LMFIT)
     #TODO! Should we fetch lmfit from the web or inlcude a tar.gz in the repo?
     set(LMFIT_PATCH_COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/patches/lmfit.patch)
-    FetchContent_Declare(
-        lmfit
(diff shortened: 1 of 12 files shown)


CMakeLists.txt
```diff
@@ -81,12 +81,12 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
...
     #TODO! Should we fetch lmfit from the web or inlcude a tar.gz in the repo?
-    set(lmfit_patch git apply ${CMAKE_CURRENT_SOURCE_DIR}/patches/lmfit.patch)
+    set(LMFIT_PATCH_COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/patches/lmfit.patch)
     FetchContent_Declare(
         lmfit
         GIT_REPOSITORY https://jugit.fz-juelich.de/mlz/lmfit.git
         GIT_TAG        main
-        PATCH_COMMAND ${lmfit_patch}
+        PATCH_COMMAND ${LMFIT_PATCH_COMMAND}
         UPDATE_DISCONNECTED 1
```
AliceMazzoleni99, line 90: On my System it always fails to fetch lmfit correctly with UPDATE_DISCONNECT set to 1.

include/aare/ClusterFile.hpp
```diff
@@ -117,29 +101,50 @@ class ClusterFile {
...
+    private:
+        ClusterVector<int32_t> read_clusters_with_cut(size_t n_clusters);
+        ClusterVector<int32_t> read_clusters_without_cut(size_t n_clusters);
+        ClusterVector<int32_t> read_frame_with_cut();
+        ClusterVector<int32_t> read_frame_without_cut();
+        bool is_selected(Cluster3x3 &cl);
+        Cluster3x3 read_one_cluster();
+};
 
+//TODO! helper functions that doesn't really belong here
```
AliceMazzoleni99, line 143: Moved them in a different file on my branch

include/aare/ClusterFile.hpp
```diff
@@ -117,29 +101,50 @@ class ClusterFile {
...
     /**
      * @brief Return the chunk size
      */
     size_t chunk_size() const { return m_chunk_size; }
+
+    /**
+     * @brief Set the region of interest to use when reading clusters. If set only clusters within
+     * the ROI will be read.
+     */
+    void set_roi(ROI roi);
```
AliceMazzoleni99, line 113: Are all these intentially passed by value?

src/ClusterFile.cpp
```diff
@@ -31,6 +31,18 @@ ClusterFile::ClusterFile(const std::filesystem::path &fname, size_t chunk_size,
     }
 }
 
+void ClusterFile::set_roi(ROI roi){
+    m_roi = roi;
+}
+
+void ClusterFile::set_noise_map(const NDView<int32_t, 2> noise_map){
+    m_noise_map = NDArray<int32_t, 2>(noise_map);
+}
```
AliceMazzoleni99, line 40: there are two copies in this function one by passing by value and one by calling the copy constructor and creating a temporary object. Isnt it better to pass by reference and only use copy assignment?

src/ClusterFile.test.cpp
```diff
@@ -0,0 +1,65 @@
...
+}
+
+
+TEST_CASE("Read clusters from single frame file", "[.integration]") {
+
+    auto fpath = test_data_path() / "clusters" / "single_frame_97_clustrers.clust";
+    REQUIRE(std::filesystem::exists(fpath));
+
+    ClusterFile f(fpath);
+    auto clusters = f.read_clusters(500);
```
AliceMazzoleni99, line 59: input argument n_ckusters a bit confusing? What does it indicate why are there not 500 clusters at the end?

CMakeLists.txt
erikfrojdh: It's an issue with cmake older than 3.28 which I didn't catch in CI. Added a workaround

include/aare/ClusterFile.hpp
erikfrojdh: ROI is just 4 integers and typically called once so I don't see an issue.

src/ClusterFile.cpp
erikfrojdh: NDView is a lightweight object, basically holding only the shape and a pointer to the data. (think passing std::string_view instead of const std::string&)

Using it has the advantage that we can create a view over a numpy array and don't need to create first the NDArray. If we already have an NDArray it's cheap to create a view over it. 

With move semantics we only get one copy, which we need since the ClusterFile in a multi threaded context needs to own the noise map. 

See mockup example: https://godbolt.org/z/8srx67PP6

src/ClusterFile.test.cpp
```diff
@@ -0,0 +1,65 @@
...
+    REQUIRE(clusters.size() == 97);
+
+
```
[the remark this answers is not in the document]
erikfrojdh: Need to update the docs for how to get test data...

include/aare/ClusterFile.hpp
erikfrojdh: Then I'll keep them and we make the move when you merge your branch

src/ClusterFile.test.cpp
erikfrojdh: n_clusters indicated the maximum number of clusters read, only if the file ends there will be fewer. The user can check with clusters.size. 

Happy to accept suggestions for a better name

AliceMazzoleni99: Ah oke I see. Just thought it was a bit misleading in the test. Maybe we can pass 97 as n_clusters or have three tests one that just tests simple single frame read were we pass the expected number of clusters and one where we pass a larger n_clusters and one where one does not pass enough clusters.

AliceMazzoleni99: Maybe expected_num_clusters would be a better variable name

AliceMazzoleni99 (approved):

PR #147, opened 2025-04-02
Api cluster vector

AliceMazzoleni99:


include/aare/CalculateEta.hpp
```diff
@@ -0,0 +1,120 @@
...
+    pLeft = 3,
+    pCenter = 4,
+    pRight = 5,
+    pTopLeft = 6,
+    pTop = 7,
+    pTopRight = 8
+} pixel;
+
+// TODO: maybe template this!!!!!! why int32_t????
+struct Eta2 {
```
erikfrojdh, line 29: The clusters come 99% of the time from 16 bit (ok 14bit) integers. To make sure we can hold the sum of the cluster it needs to be int32_t. But yes if the code should work with other types of clusters we need to template this.

include/aare/ClusterFileV2.hpp
```diff
@@ -1,15 +1,16 @@
 #pragma once
```
erikfrojdh, line 1: I think we should remove the ClusterFileV2 now and then think about how we want to store clusters in the future.

include/aare/ClusterVector.hpp
```diff
@@ -103,26 +103,22 @@ template <typename T, typename CoordType = int16_t> class ClusterVector {
...
     /**
      * @brief Add a cluster to the vector
-     * @param x x-coordinate of the cluster
-     * @param y y-coordinate of the cluster
-     * @param data pointer to the data of the cluster
-     * @warning The data pointer must point to a buffer of size cluster_size_x *
-     * cluster_size_y * sizeof(T)
      */
-    void push_back(CoordType x, CoordType y, const std::byte *data) {
+    void push_back(const ClusterType &cluster) {
```
erikfrojdh, line 107: This probably can be simplified to single memcpy and incrementing size

include/aare/ClusterVector.hpp
```diff
@@ -24,10 +29,11 @@ namespace aare {
...
  */
-template <typename T, typename CoordType = int16_t> class ClusterVector {
+template <typename T, uint8_t ClusterSizeX, uint8_t ClusterSizeY,
+          typename CoordType>
+class ClusterVector<Cluster<T, ClusterSizeX, ClusterSizeY, CoordType>> {
     using value_type = T;
-    size_t m_cluster_size_x;
-    size_t m_cluster_size_y;
+
     std::byte *m_data{};
```
erikfrojdh, line 37: Do we want to allocate std::byte or the cluster type we have. (before it was not known.)

erikfrojdh (commented):
Looks to be going in the right direction. Let's see how the templates impact the rest of the code.

python/src/cluster.hpp
```diff
@@ -16,149 +16,243 @@
...
+                self.push_back(cl);
+            },
+            py::arg("cluster"))
+        */
+
         .def("push_back",
-             [](ClusterVector<T> &self, int x, int y, py::array_t<T> data) {
-                //  auto view = make_view_2d(data);
-                 self.push_back(x, y, reinterpret_cast<const std::byte*>(data.data()));
+             [](ClusterVector<ClusterType> &self, const ClusterType &cluster) {
```
AliceMazzoleni99, line 92: @erikfrojdh do you know how to properly implement a push_back property - It does not automatically converts the python cluster objects to the c++ cluster object

include/aare/ClusterFinder.hpp
```diff
@@ -10,17 +10,21 @@
...
     const PEDESTAL_TYPE m_nSigma;
     const PEDESTAL_TYPE c2;
     const PEDESTAL_TYPE c3;
     Pedestal<PEDESTAL_TYPE> m_pedestal;
-    ClusterVector<CT> m_clusters;
+    ClusterVector<ClusterType> m_clusters;
+
+    static const uint8_t ClusterSizeX =
+        extract_template_arguments<ClusterType>::cluster_size_x;
+    static const uint8_t ClusterSizeY =
```
erikfrojdh, line 25: This looks like a bug

include/aare/Cluster.hpp
```diff
@@ -1,36 +1,121 @@
...
+constexpr bool is_valid_cluster =
+    std::is_arithmetic_v<T> && std::is_integral_v<CoordType> &&
+    (ClusterSizeX > 0) && (ClusterSizeY > 0);
+
+// requires clause c++20 maybe update
+template <typename T, uint8_t ClusterSizeX, uint8_t ClusterSizeY,
+          typename CoordType = int16_t,
+          typename Enable = std::enable_if_t<
+              is_valid_cluster<T, ClusterSizeX, ClusterSizeY, CoordType>>>
+struct Cluster {
```
erikfrojdh, line 30: Maybe we can simplify the specialisations. It's a bit annoying with the repeated declaration of members.

include/aare/ClusterFinder.hpp
```diff
@@ -10,17 +10,21 @@
...
+        extract_template_arguments<ClusterType>::cluster_size_x;
```
erikfrojdh, line 26: Nice trick to get the template parameter. Another maybe simpler way is to declare a static constexpr member exposing the cluster size: 

```cpp
Cluster<T, uint8_t ClusterSizeX>
Cluster{
...
public:
static constexpr auto cluster_size_x = ClusterSizeX;
};

```

erikfrojdh (commented):
A few comments but need a bit more time to look around and to discuss.

include/aare/ClusterFileV2.hpp
AliceMazzoleni99: removed it

include/aare/ClusterVector.hpp
AliceMazzoleni99: done

include/aare/ClusterVector.hpp
AliceMazzoleni99: changed it to a vector

include/aare/CalculateEta.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+template <typename T> struct Eta2 {
+    double x;
+    double y;
+    int c;
+    T sum;
+};
+
+/**
+ * @brief Calculate the eta2 values for all clusters in a Clsutervector
```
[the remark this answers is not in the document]
AliceMazzoleni99: done

include/aare/Cluster.hpp
```diff
@@ -1,36 +1,121 @@
...
-
-    int32_t sum_2x2() const{
-        std::array<int32_t, 4> total;
-        total[0] = data[0] + data[1] + data[3] + data[4];
-        total[1] = data[1] + data[2] + data[4] + data[5];
-        total[2] = data[3] + data[4] + data[6] + data[7];
-        total[3] = data[4] + data[5] + data[7] + data[8];
-        return *std::max_element(total.begin(), total.end());
-    }
+    T data[4];
```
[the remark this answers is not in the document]
AliceMazzoleni99: But ClusterSizeX is not defined?

include/aare/ClusterFinder.hpp
AliceMazzoleni99: Good idea i changed it

include/aare/ClusterFinder.hpp
AliceMazzoleni99: Why, can you elaborate it a bit?

include/aare/ClusterCollector.hpp
```diff
@@ -2,51 +2,57 @@
...
-        std::vector<ClusterVector<int>> steal_clusters(){
-            if(!m_stopped){
-                throw std::runtime_error("ClusterCollector is still running");
-            }
-            return std::move(m_clusters);
+  public:
+    ClusterCollector(ClusterFinderMT<ClusterType, uint16_t, double> *source) {
+        m_source = source->sink();
+        m_thread =
+            std::thread(&ClusterCollector::process,
```
erikfrojdh, line 41: The cluster finder has no knowledge of m_cluster (the vector of ClusterVectors) in only exposes m_sink as an output queue which any process can take elements from. In principle one could expose all queues so save one but that complicates it slightly for the user and move should be fast... 

Can discuss and draw on the whiteboard.

include/aare/ClusterVector.hpp
```diff
@@ -13,292 +14,154 @@
...
-        std::byte *ptr = m_data + 2 * sizeof(CoordType); // skip x and y
-
-        for (size_t i = 0; i < m_size; i++) {
-            sums[i] =
-                std::accumulate(reinterpret_cast<T *>(ptr),
-                                reinterpret_cast<T *>(ptr) + n_pixels, T{});
-            ptr += stride;
+        std::vector<T> sums(m_data.size());
+
+        for (size_t i = 0; i < m_data.size(); i++) {
```
erikfrojdh, line 79: Use a range based for loop?

include/aare/ClusterVector.hpp
```diff
@@ -13,292 +14,154 @@
...
+    ClusterType const *data() const { return m_data.data(); }
 
     /**
      * @brief Return a reference to the i-th cluster casted to type V
      * @tparam V type of the cluster
      */
-    template <typename V> V &at(size_t i) {
-        return *reinterpret_cast<V *>(element_ptr(i));
-    }
+    ClusterType &at(size_t i) { return m_data[i]; }
```
erikfrojdh, line 152: Should we change this to use the subscript operator instead? 
```
operator[]
```

I originally only had at since it looked a bit better when you had to give the type as a template in the call. 

In vector [] is not bounds checked and .at is so I think we could follow the same scheme.

erikfrojdh (approved):
One you changed the gain map we can merge

include/aare/ClusterVector.hpp
AliceMazzoleni99: True changed it.

include/aare/ClusterVector.hpp
AliceMazzoleni99: done

include/aare/CalculateEta.hpp
```diff
@@ -0,0 +1,170 @@
...
+                        (cl.data[cluster_center_index] +
+                         cl.data[cluster_center_index - ClusterSizeX]));
+    }
+
+    eta.c = c; // TODO only supported for 2x2 and 3x3 clusters -> at least no
+               // underyling enum class
+    return eta;
+}
+
+// Dont get why this is correct - photon center should be top right corner
```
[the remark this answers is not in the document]
AliceMazzoleni99: done

PR #153, opened 2025-04-07
function signature for push back

erikfrojdh:
This example now works:
```python
cl = Cluster3x3i(5,7,np.array((1,2,3,4,5,6,7,8,9), dtype = np.int32))
cv = ClusterVector_Cluster3x3i()
cv.push_back(cl)
```

### DIFF
--- a/python/src/cluster.hpp
+++ b/python/src/cluster.hpp
@@ -63,8 +63,11 @@ void define_cluster(py::module &m, const std::string &typestr) {
             });
 }
 
-template <typename ClusterType>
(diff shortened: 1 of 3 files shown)


python/src/cluster.hpp
```diff
@@ -63,8 +63,11 @@ void define_cluster(py::module &m, const std::string &typestr) {
             });
 }
 
-template <typename ClusterType>
+template <typename Type, uint8_t ClusterSizeX, uint8_t ClusterSizeY,
+          typename CoordType = uint16_t>
```
thattil, line 67: Maybe use CoordType insetad of hard coding uint16_t?

PR #154, opened 2025-04-08
fixed warnings and removed ambiguous read_frame

erikfrojdh:
Fixed warnings:
- unused variable in Interpolator
- Narrowing conversions uint64-->int64

Removed an ambiguous function from JungfrauDataFile
- NDarry read_frame(header&=nullptr)
- Frame read_frame()

### DIFF
--- a/src/Interpolator.cpp
+++ b/src/Interpolator.cpp
@@ -6,8 +6,8 @@ namespace aare {
 Interpolator::Interpolator(NDView<double, 3> etacube, NDView<double, 1> xbins,
                            NDView<double, 1> ybins, NDView<double, 1> ebins)
     : m_ietax(etacube), m_ietay(etacube), m_etabinsx(xbins), m_etabinsy(ybins), m_energy_bins(ebins) {
-    if (etacube.shape(0) != xbins.ssize() || etacube.shape(1) != ybins.ssize() ||
-        etacube.shape(2) != ebins.ssize()) {
+    if (etacube.shape(0) != xbins.size() || etacube.shape(1) != ybins.size() ||
+        etacube.shape(2) != ebins.size()) {
         throw std::invalid_argument(
             "The shape of the etacube does not match the shape of the bins");
     }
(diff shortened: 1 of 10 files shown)


src/Interpolator.cpp
```diff
@@ -6,8 +6,8 @@ namespace aare {
 Interpolator::Interpolator(NDView<double, 3> etacube, NDView<double, 1> xbins,
                            NDView<double, 1> ybins, NDView<double, 1> ebins)
     : m_ietax(etacube), m_ietay(etacube), m_etabinsx(xbins), m_etabinsy(ybins), m_energy_bins(ebins) {
-    if (etacube.shape(0) != xbins.size() || etacube.shape(1) != ybins.size() ||
-        etacube.shape(2) != ebins.size()) {
+    if (etacube.shape(0) != xbins.ssize() || etacube.shape(1) != ybins.ssize() ||
+        etacube.shape(2) != ebins.ssize()) {
         throw std::invalid_argument(
```
AliceMazzoleni99, line 11: At first glance i would have rather changed the shape in NDView to size_t instead of int64_t

src/JungfrauDataFile.cpp
```diff
@@ -54,6 +55,10 @@ size_t JungfrauDataFile::frame_number(size_t frame_index) {
     return read_header().framenum;
 }
 
+std::array<ssize_t, 2> JungfrauDataFile::shape() const {
+    return {static_cast<ssize_t>(rows()), static_cast<ssize_t>(cols())};
+}
```
AliceMazzoleni99, line 60: again would keep this unsigned and change shape type in NDArray to size_t

src/Interpolator.cpp
erikfrojdh: In the end I think we have to provide both. To create Numpy arrays using pybind11 we need a shape with signed integers. 

https://github.com/pybind/pybind11/blob/a28ea8ccc5b98af73a2306ca0e43bcf0c82d39d8/include/pybind11/numpy.h#L876

https://pybind11.readthedocs.io/en/stable/reference.html#_CPPv4N10memoryview11from_bufferEPv7ssize_tPKcN6detail13any_containerI7ssize_tEEN6detail13any_containerI7ssize_tEEb

They added std::ssize to C++20 so providing both might be a good idea. Question is if we need/want to have a unsigned shape in addition?

AliceMazzoleni99: Oh yes. I would only keeo std::ssize then.

erikfrojdh: I made size() signed and removed the ssize() function.

PR #159, opened 2025-04-10
Some fixes (need more testing later)

erikfrojdh:
- Change of pointer size caused out of bounds write
- UB to write to memory reserved by std::vector::reserver --> allocate dummy clusters by using resize instead
   - but now we can't reserve like we want to, need a fix. 
- format string not working, fixed


include/aare/ClusterVector.hpp
```diff
@@ -299,7 +299,7 @@ class ClusterVector<Cluster<T, ClusterSizeX, ClusterSizeY, CoordType>> {
      */
     ClusterVector(size_t capacity = 1024, uint64_t frame_number = 0)
         : m_frame_number(frame_number) {
-        m_data.reserve(capacity);
+        m_data.resize(capacity);
     }
```
AliceMazzoleni99, line 303: Maybe we keep reserve here and in the ClusterFile we resize to n_clusters after initializing ClusterVector

python/tests/test_Cluster.py
```diff
@@ -5,6 +5,12 @@
...
 
 
+def test_cluster_vector_can_be_converted_to_numpy():
+    cv = aare.ClusterVector_Cluster3x3i()
+    arr = np.array(cv, copy=False)
+    assert arr.shape == (0,)  # 4 for x, y, size, energy and 9 for the cluster data
+
+
 def test_ClusterVector(): 
     """Test ClusterVector"""
```
AliceMazzoleni99, line 15: maybe we can pus_back a cluster and check size

AliceMazzoleni99: I can do that actually

python/src/cluster.hpp
```diff
@@ -80,11 +80,12 @@ void define_cluster_vector(py::module &m, const std::string &typestr) {
                  self.push_back(cluster);
              })
 
+
         // implement push_back
         .def_property_readonly("size", &ClusterVector<ClusterType>::size)
         .def("item_size", &ClusterVector<ClusterType>::item_size)
         .def_property_readonly("fmt",
-                               [typestr]() { return fmt_format<ClusterType>; })
+                               [typestr](ClusterVector<ClusterType> &self) { return fmt_format<ClusterType>; })
```
AliceMazzoleni99, line 89: why do we need to pass self?

include/aare/ClusterVector.hpp
erikfrojdh: Sounds good!

python/src/cluster.hpp
erikfrojdh: Member functions are always called with the instance of the object as the first parameter. 

```python
cv = ClusterVector()
cv.fmt
```
This will fail without the self.

AliceMazzoleni99: Ah I see.

AliceMazzoleni99 (approved):

PR #161, opened 2025-04-14
Adding more tests, fixing hitmap and reading with cuts

erikfrojdh:


python/aare/ClusterFinder.py
```diff
@@ -0,0 +1,14 @@
+
+from ._aare import ClusterFinder_Cluster3x3i
+import numpy as np
+
+def ClusterFinder(image_size, cluster_size, n_sigma=5, dtype = np.int32, capacity = 1024):
+    """
```
AliceMazzoleni99, line 6: We should pass a lower capacity

AliceMazzoleni99 (approved):

PR #163, opened 2025-04-16
Custom base for decoding ADC data

erikfrojdh:
New function apply_custom_weights (can we find a better name) that takes a uint16 and a NDView<double,1> of bases for the conversion.  For each supplied weight it is used as base (instead of 2) to convert from bits to a double.

### DIFF
--- a/src/NDView.test.cpp
+++ b/src/NDView.test.cpp
@@ -191,3 +191,15 @@ TEST_CASE("compare two views") {
 
     REQUIRE((view1 == view2));
 }
+
+
+TEST_CASE("Create a view over a vector"){
+    std::vector<int> vec;
+    for (int i = 0; i != 12; ++i) {
+        vec.push_back(i);
+    }
+    auto v = aare::make_view(vec);
+    REQUIRE(v.shape()[0] == 12);
+    REQUIRE(v[0] == 0);
+    REQUIRE(v[11] == 11);
+}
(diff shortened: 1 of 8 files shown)


src/decode.test.cpp
```diff
@@ -0,0 +1,80 @@
...
+    weights_data(2) = 1.8;
+
+    auto weights = weights_data.view();
+
+
+    double output = aare::apply_custom_weights(input, weights);
+    CHECK_THAT(output, WithinAbs(1.0, 0.001));
+
+    input = 1UL << 1;
+    output = aare::apply_custom_weights(input, weights);
```
AliceMazzoleni99, line 68: Its a bit strange to assign type 1UL to uint16_t maybe a bit nicer input = 1; input = input << 1;

src/NDView.test.cpp
```diff
@@ -190,4 +190,16 @@ TEST_CASE("compare two views") {
...
 
     REQUIRE((view1 == view2));
+}
+
+
+TEST_CASE("Create a view over a vector"){
+    std::vector<int> vec;
+    for (int i = 0; i != 12; ++i) {
+        vec.push_back(i);
+    }
```
AliceMazzoleni99, line 200: only small comment but maybe for next time I really like std::iota

src/decode.test.cpp
erikfrojdh: rewrote as 1 << 1 and 1 << 2 agree looks cleaner

src/NDView.test.cpp
erikfrojdh: Then you will have it. Rewrote as:
```cpp
std::vector<int> vec(12);
std::iota(vec.begin(), vec.end(), 0);
```

PR #165, opened 2025-04-23
removed uneccecary codna numpy variants

erikfrojdh:
With numpy 2.0 we no longer need to build against every supported numpy version. This way we can save up to 6 builds. 

- https://numpy.org/doc/stable/dev/depending_on_numpy.html
- https://conda-forge.org/docs/maintainer/knowledge_base/#building-against-numpy

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -2,55 +2,49 @@ package:
   name: aare
   version: 2025.4.22 #TODO! how to not duplicate this?
 
-
-
-
-
-
 source:
   path: ..
 
 build:
   number: 0
   script:
-    - unset CMAKE_GENERATOR && {{ PYTHON }} -m pip install . -vv  # [not win]
-    - {{ PYTHON }} -m pip install . -vv  # [win]
+    - unset CMAKE_GENERATOR && {{ PYTHON }} -m pip install . -vv 
 
 requirements:
   build:
-    - python {{python}}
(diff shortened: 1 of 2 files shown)


thattil (approved):

conda-recipe/meta.yaml
```diff
@@ -2,55 +2,49 @@ package:
...
 
 requirements:
   build:
-    - python {{python}}
-    - numpy {{ numpy }}
     - {{ compiler('cxx') }}
+    - cmake
+    - ninja
+    - pip
+    - catch2
```
AliceMazzoleni99, line 19: why are ninja, pip and catch2 in build and not in host anymore

erikfrojdh: I probably misplaced them the last time: https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#build

Build should be stuff that's needed to build the package but not needed at runtime. Moved pip back to host since python is there.

AliceMazzoleni99: But shouldnt catch2 be in host as well? Its not a build tool?

erikfrojdh: On a closer look we actually don't need it at all since we just build the python pkg. 

In case we would need it I don't know if it should be in build or host... conda tends to often export the host requirements also as run and it's not needed during runtime.

AliceMazzoleni99 (approved):

PR #167, opened 2025-04-25
consistent use of ssize_t instead of int64_t

erikfrojdh:
- Consistent use of ssize_t to avoid issues on 32 bit platforms and also mac with (long long int as ssize_t)


include/aare/NumpyFile.hpp
```diff
@@ -69,7 +69,7 @@ class NumpyFile : public FileInterface {
      */
     template <typename T, size_t NDim> NDArray<T, NDim> load() {
         NDArray<T, NDim> arr(make_shape<NDim>(m_header.shape));
-        if (fseek(fp, static_cast<int64_t>(header_size), SEEK_SET)) {
+        if (fseek(fp, static_cast<long>(header_size), SEEK_SET)) {
             throw std::runtime_error(LOCATION + "Error seeking to the start of the data");
```
AliceMazzoleni99, line 73: why is this long

erikfrojdh: To match the function declaration in the c standard library: https://en.cppreference.com/w/c/io/fseek On most platforms it's the same type at int64_t but not guaranteed.

AliceMazzoleni99: :+1:

AliceMazzoleni99 (approved):

Issue #156, opened 2025-04-10
Opening RawFile fails with many subfiles


erikfrojdh:
```python
f  = RawFile('path/to/some_master_0.json')
>>> RuntimeError("Could not open ...  data file 1980")
```

Most likely linked to the fact that we open all Subfiles and keep them open. 

Solution: Refactor to keep only the current file open (like in JungfrauDataFile)

erikfrojdh:
fixed

PR #173, opened 2025-04-30
RawSubFile support multi file access

erikfrojdh:
This PR is a fix/improvement to a problem that Jonathan had. (#156) The original implementation opened all subfiles at once witch works for normal sized datasets but fails at a certain point (thousands of files). 

- This solution uses RawSubFile to manage the different file indicies and only opens the file we need
- Added logger.h from slsDetectorPackage for debug printing (in production no messages should be visible)


src/RawSubFile.cpp
```diff
@@ -130,4 +141,69 @@ size_t RawSubFile::frame_number(size_t frame_index) {
...
+    // find how many files we have and the number of frames in each file
+    m_last_frame_in_file.clear();
+    size_t file_index = m_offset;
+    
+    while (std::filesystem::exists(fpath(file_index))) {
+        auto n_frames = std::filesystem::file_size(fpath(file_index)) /
+                        (m_bytes_per_frame + sizeof(DetectorHeader));
+        m_last_frame_in_file.push_back(n_frames);
+        LOG(logDEBUG) << "Found: " << n_frames << " frames in file: " << fpath(file_index).string();
+        ++file_index;
```
AliceMazzoleni99, line 197: Dont really get this shouldnt be one frame one file? Or what does offset denote?

src/RawSubFile.test.cpp
```diff
@@ -0,0 +1,68 @@
...
+    RawSubFile f(fpath_raw, DetectorType::Jungfrau, 512, 1024, 16);
+
+
+    auto fpath_npy = test_data_path() / "raw/jungfrau" / "jungfrau_single_0.npy";
+    REQUIRE(std::filesystem::exists(fpath_npy));
+
+    //Numpy file with the same data to use as reference
+    File npy(fpath_npy, "r");
+    npy.seek(3);
+
```
AliceMazzoleni99, line 54: why frame_number 3 would have guessed that it is frame number 10 or 13?

src/RawSubFile.test.cpp
```diff
@@ -0,0 +1,68 @@
...
+    CHECK(f.frames_in_file() == 7);
+    CHECK(npy.total_frames() == 10);
+
+
+    DetectorHeader header{};
+    NDArray<uint16_t, 2> image({static_cast<ssize_t>(f.rows()), static_cast<ssize_t>(f.cols())});
+    for (size_t i = 0; i < 7; ++i) {
+        CHECK(f.tell() == i);
+        f.read_into(image.buffer(), &header);
+        CHECK(header.frameNumber == i + 4);
```
AliceMazzoleni99, line 64: why i + 4?

AliceMazzoleni99: why not + 3?

src/RawSubFile.test.cpp
erikfrojdh: The files have 3, 3, 3 and 1 frames and we start at f1 so that is why we have to seek to 3 in the numpy file. Added a comment.

src/RawSubFile.test.cpp
erikfrojdh: Frame number starts at 1 while frame index starts at 0. (i.e. 0 offset in the file). Added a comment.

src/RawSubFile.test.cpp
AliceMazzoleni99: ah got it frames_in_file denotes the number of frames in the filesystem

src/RawFile.cpp
```diff
@@ -210,70 +184,60 @@ size_t RawFile::bytes_per_pixel() const {
...
-                subfiles[subfile_id][part_idx]->frame_number(
-                    frame_index % m_master.max_frames_per_file());
+    if (n_modules() != 1) { //if we have more than one module
+        for (size_t part_idx = 0; part_idx != n_modules(); ++part_idx) {
+            frame_numbers[part_idx] = m_subfiles[part_idx]->frame_number(frame_index); 
         }
+
         // 1. if frame number vector is the same break
         while (std::adjacent_find(frame_numbers.begin(), frame_numbers.end(),
                                   std::not_equal_to<>()) !=
```
AliceMazzoleni99, line 204: maybe replace with std::equal a bit easier to read

src/RawFile.cpp
```diff
@@ -210,70 +184,60 @@ size_t RawFile::bytes_per_pixel() const {
...
             auto min_frame_idx = std::distance(
                 frame_numbers.begin(),
                 std::min_element(frame_numbers.begin(), frame_numbers.end()));
+
             // 3. increase its index and update its respective frame number
             frame_indices[min_frame_idx]++;
+
             // 4. if we can't increase its index => throw error
             if (frame_indices[min_frame_idx] >= total_frames()) {
                 throw std::runtime_error(LOCATION +
```
AliceMazzoleni99, line 217: I dont understand this part? Arent there frames or files missing if frame_number != frame_index + 1. As i understood it this can only happen if we have an offset but in RawFile we open all subfiles of a module starting at 0 so there cannot be an offset.

erikfrojdh: There is a mode in the receiver where we drop frames that are not complete. Since the receivers are independent it could end up that one file has all frames but some are missing in another file. Meaning that the frame at index 10 is a different frameNumber in the different files.

src/RawFile.cpp
erikfrojdh: I'm not sure I understand how you want to rewrite it? Can you give the proposed code?

PR #175, opened 2025-05-05
automated versioning

AliceMazzoleni99:


erikfrojdh (changes requested):
Looks good but would it be possible to remove leading 0's? For example 2025.05.03 --> 2025.5.3

erikfrojdh (approved):

PR #177, opened 2025-05-07
removed extra const on return type, added cast

erikfrojdh:
Fixed warnings on apple clang:

- removed extra const on return type
- added cast to suppress a float to double conversion warning


AliceMazzoleni99 (approved):
Looks good.

PR #182, opened 2025-05-30
Angle calibration

AliceMazzoleni99:
Some functions to ndarray and ndview  for testing the angle calibration. (Large commit history as implementation of angle claibration is moved)


thattil:
This one should also be merged from main. The base of this PR should also be changed to main, after which we should be able to archiver the developer branch

PR #192, opened 2025-06-10
Dev/fix/rawfilereader with roi

AliceMazzoleni99:
geometry is calculated from master file.


include/aare/RawMasterFile.hpp
```diff
@@ -1,5 +1,7 @@
 #pragma once
+// #include "aare/RawFile.hpp"
```
erikfrojdh, line 2: Remove commented line?

src/RawFile.cpp
```diff
@@ -67,6 +78,8 @@ void RawFile::read_into(std::byte *image_buf, size_t n_frames,
 
 size_t RawFile::n_modules() const { return m_master.n_modules(); }
 
+size_t RawFile::n_modules_in_roi() const { return m_modules_in_roi.size(); }
```
erikfrojdh, line 81: Do we need to expose this?

src/RawFile.test.cpp
```diff
@@ -160,6 +163,116 @@ TEST_CASE("Read multipart files", "[.integration]") {
     }
 }
 
+// dummy test class for easy member access
+class RawFileTestWrapper : public RawFile {
```
erikfrojdh, line 167: We should maybe have a longer discussion but I'm not sure this is the way to go. Do we need to test the implementation or is it enough to test the public interface?

src/RawMasterFile.test.cpp
```diff
@@ -145,6 +146,19 @@ TEST_CASE("Parse a master file in .json format", "[.integration]") {
     REQUIRE_FALSE(f.digital_samples());
 }
 
+TEST_CASE("Parse a master file in old .raw format",
```
erikfrojdh, line 149: I don't think we need all tags. I would settle on one... maybe swap out to .with-data? 
Selecting tests from one file can be done using -# in the command line enabling file names as tags

src/RawFile.cpp
```diff
@@ -15,11 +15,22 @@ namespace aare {
 RawFile::RawFile(const std::filesystem::path &fname, const std::string &mode)
     : m_master(fname) {
     m_mode = mode;
+
     if (mode == "r") {
         find_geometry();
         if (m_master.roi()) {
             m_geometry =
                 update_geometry_with_roi(m_geometry, m_master.roi().value());
+            m_modules_in_roi.reserve(n_modules());
```
erikfrojdh, line 24: Should this code be here or in update_geometry_with_roi?

src/RawFile.cpp
```diff
@@ -15,11 +15,22 @@ namespace aare {
...
+            for (size_t module_index = 0; module_index < n_modules();
+                 ++module_index) {
+                if (m_geometry.module_pixel_0[module_index].width != 0 &&
+                    m_geometry.module_pixel_0[module_index].height != 0)
+                    m_modules_in_roi.push_back(module_index);
+            }
+        } else {
+            m_modules_in_roi.resize(n_modules());
+            std::iota(m_modules_in_roi.begin(), m_modules_in_roi.end(), 0);
```
erikfrojdh, line 33: +1 for iota =)

include/aare/RawFile.hpp
```diff
@@ -30,11 +29,17 @@ struct ModuleConfig {
  * Consider using that unless you need raw file specific functionality.
  */
 class RawFile : public FileInterface {
+
+    friend class RawMasterFile;
```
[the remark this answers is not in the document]
AliceMazzoleni99: Yes I wanted to access RawFile::read_header() which was private/protected in another private method RawMasterFile::retrieve_geometry. I guess its better to just declare RawFile::read_header() as public.

src/RawFile.cpp
AliceMazzoleni99: well kept it for easy testing

include/aare/RawMasterFile.hpp
```diff
@@ -61,6 +63,8 @@ class ScanParameters {
     void increment_stop();
 };
 
+class RawFile; // forward declaration
```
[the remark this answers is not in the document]
AliceMazzoleni99: Removed it.

include/aare/RawFile.hpp
```diff
@@ -95,18 +98,17 @@ class RawFile : public FileInterface {
      */
     Frame get_frame(size_t frame_index);
 
+    void open_subfiles();
 
+  protected:
```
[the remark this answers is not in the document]
AliceMazzoleni99: Implemented a friend_test macro probably should have been in another PR

erikfrojdh:
@AliceMazzoleni99 @thattil I'm more or less convinced that we should have a way to also test private implementations, but I'm not sure that it's needed here. Most properties are anyway visible from the outside. Let's discuss quickly over coffee...

PR #210, opened 2025-06-27
naive implementation of 3x3 and 5x5 reduction

erikfrojdh:
Can this be generalized?


AliceMazzoleni99 (approved):
General 3x3, 2x2 reduction implemented.

PR #219, opened 2025-07-25
Move constructor to drop 1st dimension of NDArray

erikfrojdh:
- helper function to initialize shape
- helper function to calculate the number of elements
- move constructor to create a NDArray<T, Ndim-1> if sizes match

### DIFF
--- a/include/aare/NDArray.hpp
+++ b/include/aare/NDArray.hpp
@@ -202,7 +202,6 @@ class NDArray : public ArrayExpr<NDArray<T, Ndim>, Ndim> {
     const T &operator[](ssize_t i) const { return data_[i]; }
 
     T *data() { return data_; }
-    T *&data_ref() { return data_; }
     std::byte *buffer() { return reinterpret_cast<std::byte *>(data_); }
     ssize_t size() const { return static_cast<ssize_t>(size_); }
     size_t total_bytes() const { return size_ * sizeof(T); }
--- a/include/aare/NDArray.hpp
+++ b/include/aare/NDArray.hpp
@@ -25,7 +25,7 @@ template <typename T, ssize_t Ndim = 2>
 class NDArray : public ArrayExpr<NDArray<T, Ndim>, Ndim> {
     std::array<ssize_t, Ndim> shape_;
     std::array<ssize_t, Ndim> strides_;
-    size_t size_{};
+    size_t size_{}; //TODO! do we need to store size when we have shape?
     T *data_;
 
   public:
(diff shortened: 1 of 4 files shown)


include/aare/NDArray.hpp
```diff
@@ -79,6 +78,24 @@ class NDArray : public ArrayExpr<NDArray<T, Ndim>, Ndim> {
...
+    //Move constructor from an an array with Ndim + 1
+    template <ssize_t M, typename = std::enable_if_t<(M == Ndim + 1)>>
+    NDArray(NDArray<T, M> &&other) 
+        : shape_(drop_first_dim(other.shape())),
+          strides_(c_strides<Ndim>(shape_)), size_(num_elements(shape_)),
+          data_(other.data()) {
+
+            // For now only allow move if the size matches, to avoid unreachable data
+            // if the use case arises we can remove this check
+            if(size() != other.size()) {
```
AliceMazzoleni99, line 91: Ah maybe I would call reset() as well and not just devalidate data_

include/aare/NDArray.hpp
```diff
@@ -25,7 +25,7 @@ template <typename T, ssize_t Ndim = 2>
 class NDArray : public ArrayExpr<NDArray<T, Ndim>, Ndim> {
     std::array<ssize_t, Ndim> shape_;
     std::array<ssize_t, Ndim> strides_;
-    size_t size_{};
+    size_t size_{}; //TODO! do we need to store size when we have shape?
     T *data_;
```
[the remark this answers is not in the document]
erikfrojdh: I tend to agree, but we need to see where it's used. Will leave it to another PR.

include/aare/NDArray.hpp
erikfrojdh: If we throw from the constructor we won't have any object so I don't think reset is needed. In the destructor it will try to delete data_, which is a nullptr so that is fine. The other object will deallocate it's own data.

include/aare/NDView.hpp
```diff
@@ -26,6 +26,32 @@ Shape<Ndim> make_shape(const std::vector<size_t> &shape) {
...
+ * @param shape The shape to drop the first dimension from.
+ * @return A new shape with the first dimension dropped.
+ */
+template<size_t Ndim>
+Shape<Ndim-1> drop_first_dim(const Shape<Ndim> &shape) {
+    Shape<Ndim - 1> new_shape;
+    std::copy(shape.begin() + 1, shape.end(), new_shape.begin());
+    return new_shape;
+}
+
```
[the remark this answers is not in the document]
erikfrojdh: Using static_assert to get a a better error message.

include/aare/NDArray.hpp
AliceMazzoleni99: Ah the other.reset is never reached got it now

PR #223, opened 2025-08-19
General reduce

AliceMazzoleni99:
Generalized reduction to 3x3 and 3x3 clusters for general sized clusters.


erikfrojdh (changes requested):
Check that the reductions work from python

PR #225, opened 2025-08-22
induce the cluster size of ClusterCollector from ClusterFinderMT - ha…

AliceMazzoleni99:
In ClusterCollector induces cluster size from passed ClusterFinderMT.


erikfrojdh (changes requested):
If the cluster collection is only used with the ClusterFinderMT maybe we always want to infer the right cluster size and take away the option to specify it in python?

AliceMazzoleni99:
> If the cluster collection is only used with the ClusterFinderMT maybe we always want to infer the right cluster size and take away the option to specify it in python?

Yes I agree. I kept it due to backwards compatibility but I can remove it.

PR #227, opened 2025-09-04
Fix: make ProducerConsumerQueue move-safe

kferjaoui:


erikfrojdh:
Closing this one as I think removing the move operators is a better way to go. #383 
Nothing in aare needs to move the queue at this stage.

PR #231, opened 2025-10-08
Fix/adapt and test interpolation

AliceMazzoleni99:
Adapted eta interpolation: 

## Eta Calculation: 
- previously assumed photon hit to be in bottom left pixel of cluster (photon hit assumed in bottom right pixel of cluster) 
-  clusters are filled from top left to bottom right (previously assumed: bottom left to top right)

## Actual Interpolation: 
- photon hits are given in pixel coordinates (previous interpolation assumed euclidean coordinates, e.g. positive distance in y coordinate becomes negative distance in row pixels) 
- removed *2 of calculated distance 

## General Adaption: 
- max_sum_2x2 return subcluster index relative to cluster center e.g. bottomleft, bottomright 

## Added proper test case 
- simulated photon hit with normal energy distribution 
- Note: Test case for 2x2 cluster fails - Think uniform photon hit distribution cant be modeled by normalized eta distribution for 2x2 clusters

### DIFF
--- a/etc/dev-env.yml
+++ b/etc/dev-env.yml
@@ -13,4 +13,5 @@ dependencies:
   - pybind11
   - numpy
   - matplotlib
+  - nlohmann_json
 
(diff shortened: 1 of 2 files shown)


erikfrojdh (approved):
Looks good, I think you can merge. We should however add integration tests based on the simulation data as soon as possible.

PR #267, opened 2026-01-19
Dev/automate tests using data

AliceMazzoleni99:
- automatically run python tests 
- automatically run test using data files on local runner from gitea
- fixed some of the workflows


erikfrojdh (approved):
Looks good to me. We need to keep an eye on the size of the test  data repo since it's now downloaded each time but at the moment it only takes ~60s. 

One could put the requirements for the tests in a separate file to reduce duplicated code but not high priority

PR #286, opened 2026-03-10
Dev/strixels/remap_simple

vhinger182:
Refactor of PR #275, branching off dev/strixels/remap
The structure has been simplified.

Basic structure:
- defs -> pure data structures (geometry, config structs, placement, remapping result)
- config -> hardcoded sensor definitions (e.g. Jungfrau iLGAD multipitch, 25um pitch group, ...)
- algo -> generic remapping algorithms
- generate -> user-facing convenience functions (i.e. generate a map for Jungfrau iLGAD multipitch sensor 25um pitch group)

So far, only a few basic sensor configurations are implemented. But in principle, it should be straight forward to implement a new one by
1. Adding geometry definitions in config
2. Adding generate function

We could also discuss a further simplification making the sensor definitions types, i.e. instead of
```
auto map = jungfrau_ilgad_singlechip_25um_strixel_map(roi, sensor_placement);
```
we template on the sensor type
```
auto map = generate_maps<Jungfrau_iLGAD_MP25>(roi, sensor_placement);
```
Then adding a new sensor just requires a new type definition.


include/aare/InclusiveROI.hpp
```diff
@@ -0,0 +1,135 @@
...
+}
+static inline InclusiveROI to_local(InclusiveROI const &roi) {
+    return {0, roi.xmax - roi.xmin, 0, roi.ymax - roi.ymin};
+}
+
+// mirroring (with respect to external grid given by width and height)
+static inline InclusiveROI mirrorX(InclusiveROI r, ssize_t width) {
+    int x0p = (width - 1) - r.xmax;
+    int x1p = (width - 1) - r.xmin;
+    return {x0p, x1p, r.ymin, r.ymax};
```
AliceMazzoleni99, line 98: not sure what it should do: shouldnt it be width + width - r.xmax

include/aare/InclusiveROI.hpp
```diff
@@ -0,0 +1,135 @@
...
+static inline InclusiveROI intersect(InclusiveROI const &a,
+                                     InclusiveROI const &b) {
+    InclusiveROI r;
+    r.xmin = std::max(a.xmin, b.xmin);
+    r.ymin = std::max(a.ymin, b.ymin);
+    r.xmax = std::min(a.xmax, b.xmax);
+    r.ymax = std::min(a.ymax, b.ymax);
+
+    if (r.xmin > r.xmax || r.ymin > r.ymax) {
+        std::cout << "WARNING: ROIs do not intersect!" << std::endl;
```
AliceMazzoleni99: you can use LOG(TLogLevel::logWARNING)) << "ROIs do not intersect!" << std::endl; this already takes care of coloring and type of message

include/aare/InclusiveROI.hpp
```diff
@@ -0,0 +1,135 @@
...
+}
+
+static inline InclusiveROI alignROIs(InclusiveROI const &roi_user,
+                                     InclusiveROI const &roi_base) {
+    const int dx = roi_base.xmin;
+    const int dy = roi_base.ymin; // + bond_shift_y;
+
+    return {roi_user.xmin - dx, roi_user.xmax - dx, roi_user.ymin - dy,
+            roi_user.ymax - dy};
+    // return translate(roi_user, roi_base.xmin, roi_base.ymin);
```
AliceMazzoleni99: not sure what this should do. Generally add documentation everywhere. You could use translate(roi_user, -roi_base.xmin, -roi_base.ymin) or should it be return {roi_base.xmin, roi_user.xmax - dx, roi_base.ymin,
            roi_user.ymax - dy}.

include/aare/RemapAlgorithm.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+namespace aare::remap::algo {
+
+constexpr defs::Rotation flip(defs::Rotation r) noexcept {
+    switch (r) {
+    case defs::Rotation::Normal:
+        return defs::Rotation::Inverse;
+    case defs::Rotation::Inverse:
+        return defs::Rotation::Normal;
+    }
```
AliceMazzoleni99: maybe add default: and throw a runtime error

include/aare/RemapDefs.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+#include "aare/InclusiveROI.hpp"
+#include "aare/NDArray.hpp"
+
+#include <vector>
+
+namespace aare::remap::defs {
+
+// Orientation of the sensor with respect to the ASICs / the HDI
+// Normal means lower part of sensor (as in GDS) aligns with lower part of HDI
```
AliceMazzoleni99: maybe add better docstring - is inverse rotated by 90 degrees or 180 ?

include/aare/RemapConfig.hpp
```diff
@@ -0,0 +1,283 @@
...
+#include "aare/RemapDefs.hpp"
+
+namespace aare::remap::config::jungfrau {
+
+/************************************
+ * Default strixel geometries
+ ************************************/
+inline constexpr defs::SensorStrixelGeometry StrxP25{.multiplicity = 3,
+                                                     .pitch_um = 25.0};
+inline constexpr defs::SensorStrixelGeometry StrxP15{.multiplicity = 5,
```
AliceMazzoleni99: ah nice didnt know that. But designated initialization is part of c++20. Although clang and gcc allow it I dont think we should use it or it breaks portability as we are using c++17. But maybe we can ask Dhanya or Erik.

include/aare/RemapConfig.hpp
```diff
@@ -0,0 +1,283 @@
...
+/************************************
+ * Single chip, multi-pitch, iLGAD
+ ************************************/
+inline constexpr defs::SensorPixelGeometry SingleChipMP_iLGAD_pix{
+    .num_pix_x = 256, .num_pix_y = 256, .guardring = {.x = 9, .y = 9}};
+
+inline constexpr defs::SensorGroupConfig SingleChipMP_iLGAD_P25{
+    .pixel = SingleChipMP_iLGAD_pix,
+    .strixel = StrxP25,
+    .placement_on_sensor = {SingleChipMP_iLGAD_pix.guardring.x + 1,
```
AliceMazzoleni99: not sure does indexing always start at zero? e.g. guardring of 9 pixels then roi starts at pixel 9 for zero indexing? Maybe I misunderstand something.

include/aare/RemapConfig.hpp
```diff
@@ -0,0 +1,283 @@
...
+    .pixel = SingleChipMP_iLGAD_pix,
+    .strixel = StrxP15,
+    .placement_on_sensor = {SingleChipMP_iLGAD_pix.guardring.x + 3,
+                            SingleChipMP_iLGAD_pix.num_pix_x -
+                                SingleChipMP_iLGAD_pix.guardring.x - 1,
+                            SingleChipMP_iLGAD_pix.num_pix_y / 4,
+                            (SingleChipMP_iLGAD_pix.num_pix_y / 4) * 2 - 1}};
+
+inline constexpr defs::SensorGroupConfig SingleChipMP_iLGAD_P18{
+    .pixel = SingleChipMP_iLGAD_pix,
```
AliceMazzoleni99: Maybe this pixel grid information should belong to struct SensorConfig?

include/aare/RemapConfig.hpp
```diff
@@ -0,0 +1,283 @@
...
+    .strixel = StrxP25,
+    .placement_on_sensor = {
+        Quad_iLGAD_pix.guardring.x + 2,
+        Quad_iLGAD_pix.num_pix_x - Quad_iLGAD_pix.guardring.x - 1,
+        Quad_iLGAD_pix.guardring.y, (Quad_iLGAD_pix.num_pix_y / 2) - 2}};
+} // namespace aare::remap::config::jungfrau
+
+/**************************************
+ * Legacy for reference, to be deleted
+ **************************************/
```
AliceMazzoleni99: yes delete all of this

include/aare/RemapFormat.hpp
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include "aare/RemapDefs.hpp"
+
+namespace aare::remap::format {
+static inline std::string toString(defs::Rotation);
+static inline std::string toString(defs::SensorPixelGeometry const &g);
+static inline std::string toString(defs::SensorStrixelGeometry const &g);
+static inline std::string toString(defs::SensorGroupConfig const &c);
```
AliceMazzoleni99: Are the conversions every used? Or only used for debugging? You could easily use , It doesnt hurt to keep it but I would move everything to to_string.hpp and maybe use lower case to_string to keep consistent with the rest of the code.

include/aare/RemapAlgorithm.hpp
```diff
@@ -0,0 +1,59 @@
...
+// Is it better to pass defs::SensorGroupConfig const& and return a copy?
+void apply_rotation_shift(defs::SensorGroupConfig &cfg,
+                          defs::BondShift bond_shift, defs::Rotation rot);
+
+defs::StrixelGroupToPixelMap strixel_to_pixel_map(
+    defs::SensorGroupConfig const &, defs::SensorPlacement const &,
+    InclusiveROI const &user_roi, defs::BondShift bond_shift = {0, 0});
+
+std::vector<defs::StrixelGroupToPixelMap> strixel_to_pixel_maps(
+    defs::SensorConfig const &, std::vector<defs::SensorPlacement> const &,
```
AliceMazzoleni99: why are there multiple sensorplacements? I thought SensorConfig is unique for one sensor and thus only one placement on the module?

include/aare/RemapDefs.hpp
```diff
@@ -0,0 +1,68 @@
...
+    InclusiveROI placement_on_sensor; // location of the group in the sensor
+                                      // coordinate system (or reduced roi)
+};
+
+// Combine multiple strixel areas that make up a single sensor
+struct SensorConfig {
+    std::vector<SensorGroupConfig> group_configs;
+};
+
+// Define location and orientation of a sensor on the module
```
AliceMazzoleni99: maybe SensorModulePlacement is more explicit?

include/aare/RemapAlgorithm.hpp
```diff
@@ -0,0 +1,59 @@
...
+    for (size_t row = 0; row < order_map.shape(0); ++row) {
+        for (size_t col = 0; col < order_map.shape(1); ++col) {
+            auto flat_index = order_map(row, col);
+            if (flat_index >= 0 &&
+                static_cast<size_t>(flat_index) < input.size()) {
+                T const &value = input[flat_index];
+                output(row, col) = value;
+                // output(row, col) = static_cast<T>(input[flat_index]);
+            } else {
+                output(row, col) = static_cast<T>(0); // or nan?
```
AliceMazzoleni99: I would throw a runtime error?

include/aare/RemapAlgorithm.hpp
```diff
@@ -0,0 +1,59 @@
...
+/**
+ *  Public API:
+ *  Applies a given remapping rule to an input array.
+ *
+ * \param input Original array
+ * \param order_map Rule for remapping
+ * \param output Remapped array
+ */
+template <typename T>
+void ApplyRemap(NDView<T, 2> const &input, NDArray<ssize_t, 2> const &order_map,
```
AliceMazzoleni99: I would pass a view instead of reference to NDArray

AliceMazzoleni99: Maybe its better to directly return the remapped array than passing?

include/aare/StrixelPixelRemapAlgorithm.hpp
```diff
@@ -0,0 +1,59 @@
...
+combine_maps(std::vector<defs::StrixelGroupToPixelMap> const &,
+             std::vector<int> const &);
+
+/**
+ *  Public API:
+ *  Applies a given remapping rule to an input array.
+ *
+ * \param input Original array
+ * \param order_map Rule for remapping
+ * \param output Remapped array
```
AliceMazzoleni99, line 22: We could move this function to PixelMap.hpp

include/aare/RemapAlgorithm.hpp
```diff
@@ -0,0 +1,59 @@
...
+                NDArray<T, 2> &output) {
+    for (size_t row = 0; row < order_map.shape(0); ++row) {
```
AliceMazzoleni99: compiler warning: use ssize_t

AliceMazzoleni99: also for col

src/RemapAlgorithm.cpp
```diff
@@ -0,0 +1,231 @@
+#include "aare/RemapAlgorithm.hpp"
+
+#include <algorithm>
+
+namespace aare::remap::algo {
+
+// Is it better to pass defs::SensorGroupConfig const& and return a copy?
+void apply_rotation_shift(defs::SensorGroupConfig &cfg,
+                          defs::BondShift bond_shift, defs::Rotation rot) {
```
AliceMazzoleni99: yes everything should be const here

src/RemapAlgorithm.cpp
```diff
@@ -0,0 +1,231 @@
...
+    // coordinates is not a multiple of strixel ncols)
+    if (group_config.placement_on_sensor.width() % multiplicity != 0)
+        throw std::logic_error("Group ROI width not divisible by multiplicity");
+
+    const int tot_ncols_strx =
+        group_config.placement_on_sensor.width() / multiplicity;
+
+    // Define mod ordering (Normal or Inverse)
+    std::vector<int> mods(multiplicity);
+    for (int i = 0; i < multiplicity; i++)
```
AliceMazzoleni99: maybe use std::iota

src/RemapAlgorithm.cpp
```diff
@@ -0,0 +1,231 @@
...
+        mods[i] = i;
+    if (rot == defs::Rotation::Inverse)
+        std::reverse(mods.begin(), mods.end());
+
+    // -- 1) Transform user roi (rx_roi) into sensor-local coordinates
+    InclusiveROI roi_user_local =
+        inclusiveroi::geom::alignROIs(roi_user, placement.placement_on_module);
+    std::cout << "Transformed user ROI: " << roi_user_local << std::endl;
```
AliceMazzoleni99: remove std::cout or add LOG(logDEBUG) <<

src/RemapAlgorithm.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+    // DEBUG
+    std::cout << "DEBUG: Group ROI before transformation (as in global config)"
```
AliceMazzoleni99: also replace by LOG(logDEBUG)

include/aare/RemapDefs.hpp
```diff
@@ -0,0 +1,68 @@
...
+struct SensorPlacement {
+    InclusiveROI placement_on_module; // location of the sensor in full-module
+                                      // coordinate system
+    Rotation rotation;
+};
+
+// Possibly, this may need to be moved to a dedicated "remapping" class
+struct StrixelGroupToPixelMap {
+    int multiplicity;
+    double pitch_um;
```
AliceMazzoleni99: yes I would remove these things. Is placement_on_sensor the same ans in SensorGroupConfig?

src/RemapAlgorithm.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+    return maps;
+}
+
+defs::StrixelGroupToPixelMap
+combine_maps(std::vector<defs::StrixelGroupToPixelMap> const &maps,
+             std::vector<int> const &gaps) {
+
+    if (maps.size() != gaps.size()) {
+        throw std::logic_error("Gaps provided are inclompatible with "
```
AliceMazzoleni99: typo: incompatible

src/RemapAlgorithm.cpp
```diff
@@ -0,0 +1,231 @@
...
+            throw std::logic_error("Maps contain incompatible multiplicities.");
+        }
+
+        if (maps[i].pitch_um != p) {
+            throw std::logic_error("Maps contain incompatible pitches.");
+        }
+
+        auto [temp_rows, temp_cols] = maps[i].map.shape();
+        global_cols = std::max(global_cols, temp_cols);
+        global_rows = global_rows + temp_rows + gaps[i];
```
AliceMazzoleni99: not sure what gaps[i] should be due to placement on sensor?

src/RemapAlgorithm.cpp
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    NDArray<ssize_t, 2> map({global_rows, global_cols}, -1);
+
+    for (size_t i = 0; i < maps.size(); ++i) {
+        auto [rows, cols] = maps[i].map.shape();
+
+        // DEBUG
+        std::cout << "DEBUG: Row offset: i = " << i
```
AliceMazzoleni99: again LOG(logDEBUG) <<

src/RemapAlgorithm.cpp
```diff
@@ -0,0 +1,231 @@
...
+        if (maps[i].pitch_um != p) {
+            throw std::logic_error("Maps contain incompatible pitches.");
+        }
+
+        auto [temp_rows, temp_cols] = maps[i].map.shape();
+        global_cols = std::max(global_cols, temp_cols);
+        global_rows = global_rows + temp_rows + gaps[i];
+    }
+
+    NDArray<ssize_t, 2> map({global_rows, global_cols}, -1);
```
AliceMazzoleni99: skip -1 in Applyremap?

include/aare/StrixelPixelRemapGenerate.hpp
```diff
@@ -0,0 +1,87 @@
+#pragma once
+
```
AliceMazzoleni99, line 2: maybe better to remap the file to StrixelPixelMap.hpp

AliceMazzoleni99: maybe its better to rename all files to StrixelPixelRemap... makes it more explicit what it is

include/aare/StrixelPixelRemapGenerate.hpp
```diff
@@ -0,0 +1,87 @@
...
+inline defs::StrixelGroupToPixelMap
+jungfrau_ilgad_singlechip_18um_strixel_map(InclusiveROI rx_roi,
+                                           defs::SensorPlacement placement,
+                                           defs::BondShift bs = {0, 0}) {
+    return algo::strixel_to_pixel_map(config::jungfrau::SingleChipMP_iLGAD_P18,
+                                      placement, rx_roi, bs);
+};
+
+inline std::vector<defs::StrixelGroupToPixelMap>
+jungfrau_ilgad_singlechip_multipitch_strixel_maps(InclusiveROI rx_roi,
```
AliceMazzoleni99, line 238: I would make this independant of chip_id and have a constexpr SensorConfig jungfrau_ilgad{{config::jungfrau::SingleChipMP_iLGAD_P25,
                                config::jungfrau::SingleChipMP_iLGAD_P15,
                                config::jungfrau::SingleChipMP_iLGAD_P18}}

include/aare/RemapGenerate.hpp
```diff
@@ -0,0 +1,87 @@
...
+jungfrau_ilgad_quad_25um_strixel_maps(InclusiveROI rx_roi,
+                                      defs::SensorPlacement placement) {
+
+    defs::SensorConfig configs{
+        {config::jungfrau::Quad_iLGAD_half, config::jungfrau::Quad_iLGAD_half}};
+
+    std::vector<defs::SensorPlacement> placements(configs.group_configs.size(),
+                                                  placement);
+
+    placements[1].rotation = algo::flip(placement.rotation);
```
AliceMazzoleni99: what does flip do?

include/aare/StrixelPixelRemapGenerate.hpp
```diff
@@ -0,0 +1,87 @@
...
+        {config::jungfrau::Quad_iLGAD_half, config::jungfrau::Quad_iLGAD_half}};
+
+    std::vector<defs::SensorPlacement> placements(configs.group_configs.size(),
+                                                  placement);
+
+    placements[1].rotation = algo::flip(placement.rotation);
+
+    return algo::strixel_to_pixel_maps(configs, placements, rx_roi);
+}
+
```
AliceMazzoleni99, line 504: functions definitly need to be commented. Maybe we only need to define one jungfrau pixel map with the combined maps? If someone wants the sensor group map one has still the freedom to calculate it.

AliceMazzoleni99 (commented):
Added a few minor comments. But it needs to be documented more precisely. Every function in particular the functions in RemapAlgorithm and RemapGenerate need to have a proper docstring e.g.
`
/** @brief short explanation
* @param define all parameters
* @return what does it return
* / 
`
and all the struct members should also have a /// @brief short member explanation 
Otherwise it looks very good.

include/aare/InclusiveROI.hpp
vhinger182: It should mirror an ROI horizontally within a given grid. I have tried to document it more clearly in the new commit a6927310a5ed321f87f2c6d597adb11accd0768d

```
// Mirror an ROI by reflecting its x-coordinates within a grid of width
// 'width'.
//
// The grid is assumed to span x = [0, width-1].
// The ROI is translated to the horizontally mirrored position while
// preserving its size.
//
// Example:
//
//  x=0           x=width-1              x=0           x=width-1
//  ----------------------- y=height-1 -----------------------
//  |                     |            |                     |
//  |    ***************  |            |  ***************    |
//  |    *         roi *  |     ->     |  *mirrored roi *    |
//  |    *             *  |            |  *             *    |
//  |    ***************  |            |  ***************    |
//  |                     |            |                     |
//  -----------------------    y=0     -----------------------
//
static inline InclusiveROI mirrorX(InclusiveROI r, ssize_t width) {
    int x0p = (width - 1) - r.xmax;
    int x1p = (width - 1) - r.xmin;
    return {x0p, x1p, r.ymin, r.ymax};
}
```

include/aare/InclusiveROI.hpp
vhinger182: Have adapted it in a6927310a5ed321f87f2c6d597adb11accd0768d

include/aare/InclusiveROI.hpp
vhinger182: Yes, good point! I have simplified it and documented it more clearly in a6927310a5ed321f87f2c6d597adb11accd0768d

```
// Rebase an ROI into the coordinate system of another ROI.
//
// The returned ROI has the same physical extent, but is expressed relative
// to roi_base instead of the original coordinate system.
//
// In other words, roi_base.xmin/ymin become the new origin (0,0).
//
// Example:
//
//   Global coordinates:
//
//      roi_base  = [100..199] x [50..149]
//      roi_input = [120..139] x [70..89]
//
//   After rebasing:
//
//      roi_input = [20..39] x [20..39]
//
// This is equivalent to translating roi_input by
// (-roi_base.xmin, -roi_base.ymin).
static inline InclusiveROI rebaseROI(InclusiveROI const &roi_input,
                                     InclusiveROI const &roi_base) {
    return translate(roi_input, -roi_base.xmin, -roi_base.ymin);
}
```

include/aare/RemapAlgorithm.hpp
vhinger182: I would prefer not to add a default that throws. If, at one point, the enum grows, a missing default will automatically trigger a compiler warning, which is helpful. And throwing means that we expect a recoverable runtime condition that could pass an invalid enum. But an invalid enum would always point to some larger programming bug and should ideally be caught at compile time.

One could use `std::unreachable()`, but that requires C+++23 and I don't think it would be worth it.

I have added
`assert(false && "Invalid Rotation passed to flip");` in 14bfc19b6a217135d1df6b6fd887b0fa155da60d

include/aare/RemapAlgorithm.hpp
vhinger182: Yes, true, in my mind, I mixed up the SensorPlacement and the placements of the groups within a single sensor. Fixed to use one SensorPlacement for all. (14bfc19b6a217135d1df6b6fd887b0fa155da60d)

include/aare/StrixelPixelRemapAlgorithm.hpp
vhinger182: Agreed. For now, I left it here, but if everything else is green-lighted, I think we could move it.

include/aare/RemapAlgorithm.hpp
vhinger182: Agree to pass an NDView instead. I even would pass by value since the copy is cheap and it makes the intent more clear, showing that the utility only takes a snapshot of these arrays and does not change the underlying data. (14bfc19b6a217135d1df6b6fd887b0fa155da60d)

I would not return the remapped array, since this would be costly in a chain of data processing, where ApplyRemap runs on many frames in a large for loop.

include/aare/RemapAlgorithm.hpp
vhinger182: Fixed (d3e337d06197880dbe49ab2057f5f3ca99c0bdc3)

include/aare/RemapAlgorithm.hpp
vhinger182: Agreed, but then this needs to separate the cases of `flat_index<0` which marks intentionally unmapped pixels (e.g. because these are guard ring pixels that do not contain meaningful data) and actual corruption if `static_cast<size_t>(flat_index) >= input.size()` which should throw. Implemented in 1b05488173c1249a233379fed48305fbbefc3ceb.

include/aare/RemapConfig.hpp
vhinger182: Yes, better to crosscheck if we should rewrite for c++17. Of course, would be cool to move to c++20

include/aare/RemapConfig.hpp
vhinger182: Yes, this is how it should be I think, so that the guardring pixels are not included in the group.

include/aare/RemapConfig.hpp
vhinger182: Yes, true, that would be clearer.
I have moved it and also renamed the structs related to the groups to make the decoupling from the sensor grid more clear:
SensorStrixelGeometry -> GroupStrixelGeometry
SensorGroupConfig -> GroupConfig

Implemented over all files this change affects in 6b0598cd302775ed85e3879fe7223804b43a141e and 4b8317de6907c9f683bdea77b2c79fa40504ea55

include/aare/RemapConfig.hpp
vhinger182: Done (9bc4a4d6a965613aba7d49cf7ddec5692dce4326)

include/aare/RemapDefs.hpp
vhinger182: Renamed to Identity and Rotate180 (afd92ac45770631fce350682bdba5d5b652c89f8)
Improved documentation (9344cea81978538d51b8bbc070e447f6f9afc784)

include/aare/RemapDefs.hpp
vhinger182: Agreed. Implemented in 9344cea81978538d51b8bbc070e447f6f9afc784

include/aare/RemapDefs.hpp
vhinger182: I agree that multiplicity and pitch are duplicates and should not be exposed like this in this struct.
Placement_on_sensor is actually a misnomer. It is not the same as the ROI in GroupConfig, but instead represents the effective ROI that was actually mapped (after applying potential bond shifts, rotations, and intersections with the rx_roi of the input file). So, I would keep it and rename it to "effective_roi" or "mapped_roi". In the algorithm, the actual order is
1) Bond shift
2) Rotation
3) Intersection with user rx_roi
Then: GroupConfig.placement_on_sensor -> 1) + 2) + 3) -> effective_roi

Documented suggestion for refactored struct in c5ccd44a4c7f380cee0b595f318fd41d90e3d587

include/aare/RemapGenerate.hpp
vhinger182: It used to just change Rotation::Identity to Rotation::Rotate180 and vice versa, but with the newer changes it has become superflous. It has been removed in 98cdc6ba08fc99ad618d339f1745f923a943bb83

include/aare/StrixelPixelRemapGenerate.hpp
vhinger182: I added a more generic overload that lets the user supply the SensorModulePlacment. For now, I also still left the versions that take a chip_id, but we could discuss to get rid of them.

Added predefined SensorConfig in 98cdc6ba08fc99ad618d339f1745f923a943bb83

include/aare/StrixelPixelRemapGenerate.hpp
vhinger182: Added documentation in 13f8c9ea806f9c8c187a5bb30b7983dcb6a4c0b1
There are some generators we could discuss not to expose, e.g. individual quad halves and even the vector version of it. For now, I left them in mostly to make the mechanism of the remapping explicit (i.e. that it remaps separate sections independently and then combines them into a full map).

src/RemapAlgorithm.cpp
vhinger182: I fully got rid of this utility in 98cdc6ba08fc99ad618d339f1745f923a943bb83 after replacing it with `inline InclusiveROI shift_rotate_roi` in afd92ac45770631fce350682bdba5d5b652c89f8

src/RemapAlgorithm.cpp
vhinger182: This function has been removed and replaced by detail implementation in RemapGenerate.hpp
`aare::remap::generate::detail::combine_group_maps` in 98cdc6ba08fc99ad618d339f1745f923a943bb83

src/RemapAlgorithm.cpp
vhinger182: Function deprecated since 98cdc6ba08fc99ad618d339f1745f923a943bb83

src/RemapAlgorithm.cpp
vhinger182: Function deprecated since 98cdc6ba08fc99ad618d339f1745f923a943bb83

src/RemapAlgorithm.cpp
vhinger182: Function deprecated since 98cdc6ba08fc99ad618d339f1745f923a943bb83

src/RemapAlgorithm.cpp
vhinger182: Algorithm has been clarified and documented in f3c79465d5cfef38a4d9f0e91724719fd68db165

include/aare/StrixelPixelRemapConfig.hpp
```diff
@@ -0,0 +1,156 @@
+#include "aare/RemapDefs.hpp"
+
+namespace aare::remap::config::jungfrau {
+
+/************************************
+ * Default strixel geometries
+ ************************************/
+inline constexpr defs::GroupStrixelGeometry StrxP25{.multiplicity = 3,
```
AliceMazzoleni99: again C++20 feature. Although its a nice feature.

include/aare/StrixelPixelRemapConfig.hpp
```diff
@@ -0,0 +1,156 @@
...
+/* Number of strixel columns: 85
+ * Number of strixel rows: 192
+ ********************************/
+
+inline constexpr defs::GroupConfig SingleChipMP_TEW_P15{
+    .strixel = StrxP15,
+    .placement_on_sensor = {SingleChipMP_TEW_pix.guardring.x + 1, // 1
+                            SingleChipMP_TEW_pix.num_pix_x -
+                                SingleChipMP_TEW_pix.guardring.x - 1, // 255
+                            SingleChipMP_TEW_pix.num_pix_y / 4,       // 64
```
AliceMazzoleni99: Are these hardcoded numbers e.g. /4 known from GroupStrixelGeometry?

include/aare/InclusiveROI.hpp
vhinger182: Rewrote mirroring function to make intent clearer: Mirroring on an axis (13b9f288a074a462d15231a50727114f6d0a8ca8)

include/aare/RemapFormat.hpp
vhinger182: Renamed and commented superfluous functions in 605c28bc19cbb676d2fe29e5f47bf83825d0a295

include/aare/StrixelPixelRemapGenerate.hpp
vhinger182: Filenames updated in 794275fa62cb3072636f70692665f3a0699a1e1e
CMakeLists updated in e6a42bbda75fa0544b06615eadfe266a1c3478c3

include/aare/StrixelPixelRemapConfig.hpp
vhinger182: I made a commit that replaces the designated initializers with aggregate initialization in member definition order: a0a8fb63413a91399f412ddc080433f043e30356
But have left the designated initializers as comments (for documentation and in case we decide it is time to move to C++20 or higher).

include/aare/StrixelPixelRemapConfig.hpp
vhinger182: No, unfortunately not really. It is sensor design specific that e.g. this multipitch strixel sensor is split into four sections with different layout.

For example, a multipitch sensor is split into four parts (where 1 part has geometry StrxP25, one has StrxP15, and two have StrxP18). But a Quad sensor is entirely made up of StrxP25, but for the top half of the sensor the ModuloOrdering is reversed.

include/aare/StrixelPixelRemapGenerate.hpp
```diff
@@ -0,0 +1,535 @@
...
+ * @param gap_rows Number of invalid strixel rows inserted between groups.
+ *
+ * @return Combined strixel-to-pixel map.
+ *
+ * @throws std::invalid_argument if gap_rows is negative.
+ * @throws std::logic_error if group maps have different widths.
+ */
+defs::StrixelGroupToPixelMap
+combine_group_maps(defs::StrixelGroupToPixelMap const &first,
+                   defs::StrixelGroupToPixelMap const &second, int gap_rows) {
```
AliceMazzoleni99: is size_t the better option for gap_rows

src/StrixelPixelRemapAlgorithm.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+            auto [row, col] = pixel_to_strixel(x, y);
+
+            const int map_col = col - min_col;
+            const int map_row = row - min_row;
+
+            // index into !!!ORIGINAL USER ROI GRID!!! (use local
+            // coordinates)
+            const ssize_t user_pixel =
+                static_cast<ssize_t>(y - roi_user_local.ymin) *
```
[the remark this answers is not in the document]
AliceMazzoleni99: And as I understood it - the pixel map is given in sensor local coordinate system?

src/StrixelPixelRemapAlgorithm.cpp
```diff
@@ -0,0 +1,241 @@
...
+    // LOG(logDEBUG)
+    std::cout << "aare::remap::algo::strixel_to_pixel_map: Group ROI after "
+                 "transformation (as in local transformation) "
+              << roi_group << '\n';
+
+    // -- 3) Compute effective ROI = intersection( roi_user, roi_group )
+    // Only pixels covered by both the user ROI and the transformed group
+    // contribute to this map.
+    const InclusiveROI effective_roi =
+        inclusiveroi::geom::intersect(roi_user_local, roi_group);
```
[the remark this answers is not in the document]
vhinger182: Should this be printed also with LOG(logDEBUG)? Or another level?

src/StrixelPixelRemapAlgorithm.cpp
```diff
@@ -0,0 +1,241 @@
...
+    // Determine the ordering of strixels within each multiplicity group.
+    std::vector<int> mods(multiplicity);
+    std::iota(mods.begin(), mods.end(), 0);
+
+    if (group_config.routing.mod_order == defs::ModuloOrdering::Reverse)
+        std::reverse(mods.begin(), mods.end());
+
+    // -- 1) Rebase the user ROI (rx_roi) into sensor-local coordinates
+    const InclusiveROI roi_user_local =
+        inclusiveroi::geom::rebaseROI(roi_user, placement.placement_on_module);
```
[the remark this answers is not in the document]
vhinger182: It should also work if the ROI is larger than the sensor.

src/StrixelPixelRemapAlgorithm.cpp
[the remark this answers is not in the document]
vhinger182: No, I think it has to be `y - roi_user_local.ymin` because we want to transform the output back into the original coordinate system of the user-supplied ROI.
For the remapping, we are working in sensor-local coordinates, meaning we are telling the algorithm in sensor-local coordinates which group of pixels to remap. However, the returned map points to a pixel in the original coordinate system (which, in the worst case could be a module-sized 512x1024 pixel grid). We need to map correctly into this reference grid by returning a flattened pixel index that refers to the original, overlying, user-supplied pixel ROI.

Coordinate systems of `map(map_row, map_col) = user_pixel`:
`map_row` and `map_col`: A local strixel coordinate system, starting at (0,0)
`user_pixel`: Flat pixel index within the original user-supplied pixel grid, typically given by `rx_roi` of the underlying data file(s).

include/aare/StrixelPixelRemapGenerate.hpp
vhinger182: I changed it to size_t in d7addfba3e323ca5169166ac0c36ffcaef920e49 and removed the negative-value check since that is now explicitly not allowed by the type.

src/StrixelPixelRemapAlgorithm.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+            auto [row, col] = pixel_to_strixel(x, y);
+
+            if (col >= total_strixel_columns)
+                continue;
+
+            min_row = std::min(min_row, row);
+            max_row = std::max(max_row, row);
+            min_col = std::min(min_col, col);
+            max_col = std::max(max_col, col);
```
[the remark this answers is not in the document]
vhinger182: I was not 100% sure the direct calculation of min and max would always work. But I agree, it should be always unambiguous. Simplified this section in 142ffad0794a962030129ae138f2d3fa9acd325a

src/StrixelPixelRemapAlgorithm.cpp
```diff
@@ -0,0 +1,241 @@
...
+    std::iota(mods.begin(), mods.end(), 0);
+
+    if (group_config.routing.mod_order == defs::ModuloOrdering::Reverse)
+        std::reverse(mods.begin(), mods.end());
+
+    // -- 1) Rebase the user ROI (rx_roi) into sensor-local coordinates
+    const InclusiveROI roi_user_local =
+        inclusiveroi::geom::rebaseROI(roi_user, placement.placement_on_module);
+    // LOG(logDEBUG)
+    std::cout
```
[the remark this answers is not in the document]
vhinger182: I had it as `std::cout` for my quick checks since compiling with AARE_VERBOSE prints a massive amount of text and I did not know how to make it only print the messages I was interested in. Changed it back again in 142ffad0794a962030129ae138f2d3fa9acd325a

src/StrixelPixelRemapAlgorithm.cpp
```diff
@@ -0,0 +1,241 @@
...
+ * Bond shifts are applied before rotation.
+ * The order is intentional because bond shifts are defined in the
+ * sensor's native coordinate system.
+ */
+inline InclusiveROI shift_rotate_roi(InclusiveROI roi,
+                                     defs::SensorPixelGeometry const &pixel,
+                                     defs::BondShift bond_shift,
+                                     defs::Rotation rot) {
+    // If there is a bond shift, translate the roi
+    if (bond_shift.x != 0 || bond_shift.y != 0)
```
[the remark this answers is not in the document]
vhinger182: Changed in 142ffad0794a962030129ae138f2d3fa9acd325a

src/StrixelPixelRemapAlgorithm.cpp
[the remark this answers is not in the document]
AliceMazzoleni99: No I would say LOG(logWARNING)

src/InclusiveROI.test.cpp
```diff
@@ -0,0 +1,221 @@
...
+        CHECK(inclusiveroi::geom::rebaseROI(input, base) ==
+              InclusiveROI{10, 19, 20, 29});
+    }
+
+    SECTION("rebasing base itself produces origin") {
+        InclusiveROI base{100, 199, 200, 299};
+
+        CHECK(inclusiveroi::geom::rebaseROI(base, base) ==
+              InclusiveROI{0, 99, 0, 99});
+    }
```
AliceMazzoleni99: Just for completeness can you add a test of an input ROI that is larger and a test of an input ROI that is completeley outside?

AliceMazzoleni99 (commented):
Nice tests. Looks good to me.

src/InclusiveROI.test.cpp
vhinger182: Added in bfb8dec2b94a723e7fb9279276239de8a941bf19

src/StrixelPixelRemapAlgorithm.cpp
```diff
@@ -160,29 +164,24 @@ strixel_to_pixel_map(defs::GroupConfig const &group_config,
...
+    // The strixel column is dx / multiplicity.
+    // Each pixel row maps onto a complete block of `multiplicity`
+    // strixel rows, regardless of the modulo ordering.
+    const int min_col = (effective_roi.xmin - roi_group.xmin) / multiplicity;
 
-    for (int y = effective_roi.ymin; y <= effective_roi.ymax; ++y) {
-        for (int x = effective_roi.xmin; x <= effective_roi.xmax; ++x) {
-
-            auto [row, col] = pixel_to_strixel(x, y);
+    const int max_col = (effective_roi.xmax - roi_group.xmin) / multiplicity;
```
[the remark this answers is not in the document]
vhinger182: The problem lies in the row calculation of pixel_to_strixel. If, by unlucky circumstance, effective_roi.ymin has a modulo!=0 then, `auto[min_row, min_col] = pixel_to_strixel(effective_roi.xmin, effective_roi.ymin)` will yield a higher value for min_row than `const int min_row = (effective_roi.ymin - roi_group.ymin) * multiplicity;` and a similar issue results for max_row if effective_roi.ymax has a modulo!=multiplicity-1.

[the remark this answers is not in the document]
vhinger182: We always want to catch the full multiplicity blocks with the bounds.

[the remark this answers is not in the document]
vhinger182: Then the strixels that are not contained by effective_roi will simply be mapped with -1. And we make sure that we do not accidentally allocate a too small array on the flip side.

src/InclusiveROI.test.cpp
vhinger182: Tomorrow, I'll work on the algorithm test

include/aare/StrixelPixelRemapDefs.hpp
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * @brief Describes the routing between ASIC pixel columns and strixel rows
+ * within a remapping group.
+ *
+ * The modulo ordering specifies whether the pixels belonging to each
+ * multiplicity group are mapped in forward or reversed order.
+ */
+struct GroupRouting {
+    ModuloOrdering mod_order = ModuloOrdering::Forward;
+};
```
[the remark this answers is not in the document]
vhinger182: Yes, in principle, we could do that. I think I originally put the struct to anticipate new future routing rules. I am not sure if this is necessary.

src/InclusiveROI.test.cpp
vhinger182: I added the test in 78862918959fffe539c652c6ed4018f0ea8cee25
Still a few test cases are missing.

vhinger182: Full test hierarchy done: 1816b0c3aba611d2a7879f720681b45251aebea0

include/aare/StrixelPixelRemapConfig.hpp
```diff
@@ -0,0 +1,173 @@
...
+     SingleChipMP_iLGAD_pix.num_pix_y - SingleChipMP_iLGAD_pix.guardring.y -
+         1}}; // 246
+/* Number of strixel columns: 59
+ * Number of strixel rows: 476
+ ********************************/
+
+// Complete SensorConfig with all groups
+inline const defs::SensorConfig SingleChipMP_iLGAD{
+    // NOTE: Vector cannot be constexpr
+    SingleChipMP_iLGAD_pix, //.pixel =
```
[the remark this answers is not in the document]
vhinger182: Yes, good point!
Done in c511231d2a88e69c99960772b91e245863722fc7

src/StrixelPixelRemapAlgorithm.cpp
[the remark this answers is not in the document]
vhinger182: Yes, horizontally on the pixel map, but they become vertical rows in the strixel map.

include/aare/StrixelPixelRemapGenerate.hpp
vhinger182: Removed in 4880b6e449deb192d4e599d1ba2bfcbff0e630f8

src/StrixelPixelRemapping/StrixelPixelRemapAlgorithm.test.cpp
```diff
@@ -0,0 +1,669 @@
...
+     *       mod = dx % multiplicity
+     *
+     *       strixel_col = dx / multiplicity
+     *       strixel_row = dy * multiplicity + mod
+     *
+     * This test deliberately calculates the expected values directly
+     * from those definitions rather than using the implementation's
+     * intermediate calculations.
+     */
+    for (ssize_t y = result.effective_roi.ymin; y <= result.effective_roi.ymax;
```
AliceMazzoleni99: can you remove that. This doesnt test anything but compares the code against itself.

src/StrixelPixelRemapping/StrixelPixelRemapAlgorithm.test.cpp
```diff
@@ -0,0 +1,669 @@
...
+
+    const auto group = test_group();
+    const auto sensor = test_sensor();
+    const auto placement = test_placement();
+
+    // User ROI is in module coordinates.
+    const InclusiveROI user_roi{60, 89, 60, 89};
+
+    // User ROI fully covers group ROI
+    const InclusiveROI user_roi_local =
```
AliceMazzoleni99: What you can do you can use inclusiveroi::geom::rebaseROI(user_roi, placement.placement_on_module); and check against hardcoded expected values.

src/StrixelPixelRemapping/StrixelPixelRemapAlgorithm.test.cpp
```diff
@@ -0,0 +1,669 @@
...
+    // The group occupies [10,39] x [10,39] in sensor-local coordinates.
+    const InclusiveROI expected_roi{10, 39, 10, 39};
+
+    CHECK(result.effective_roi == expected_roi);
+    CHECK(result.map.shape(0) == 90);
+    CHECK(result.map.shape(1) == 10);
+
+    // Rebase into sensor roi:
+    // {50, 99, 50, 99}
+    const InclusiveROI user_roi_local = {-5, 54, -5, 54};
```
AliceMazzoleni99: similar I would use rebase and check against these hardcoded. Remove the bottom part.

docs/src/StrixelPixelRemapping/index.rst
```diff
@@ -58,6 +58,32 @@ links each physical sensor strixel to its corresponding ASIC pixel. We refer to
...
+    The reference pixel grid does not necessarily have to correspond to exactly one ASIC. It could, for example, be the
+    standard output of a single JUNGFRAU detector module with dimensions :math:`2\times 4` ASICs (:math:`512\times 1024` pixels). 
+
+The API predefines known sensor configurations and provides direct map generators for these.
+
+To apply the remapping to a detector image, the API provides the following utility function:
+
+.. doxygenfunction:: aare::remap::algo::ApplyRemap
+
+TODO: Discuss groups, predefined sensors, bond shift, rotation, modulo ordering, ROIs (?) ...
```
AliceMazzoleni99: can you comment the TODO: ..

docs/figures/StrixelSketch.png
AliceMazzoleni99: can you remove this figure. I dotn think its needed.

docs/figures/StrixelRemappingConcept.png
AliceMazzoleni99: would it be possible to rename implant to strixel?

AliceMazzoleni99 (commented):
Some more comments.

include/aare/StrixelPixelRemapping/StrixelPixelRemapDefs.hpp
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * @brief Describes the routing between ASIC pixel columns and strixel rows
+ * within a remapping group.
+ *
+ * The modulo ordering specifies whether the pixels belonging to each
+ * multiplicity group are mapped in forward or reversed order.
+ */
+struct GroupRouting {
+    ModuloOrdering mod_order = ModuloOrdering::Forward;
+};
```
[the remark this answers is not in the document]
AliceMazzoleni99: Oke lets leave it as it is. In python added an overload to directly pass ModuloOrdering

src/StrixelPixelRemapping/StrixelPixelRemapAlgorithm.test.cpp
vhinger182: I have rewritten all these tests to check the values explicitly (hardcoded).

AliceMazzoleni99 (approved):

PR #300, opened 2026-04-02
bumped pybind version, cmake fixes

erikfrojdh:
- Bumped pybind11 version
- Bumped libfmt version
- Patch  Minuit to avoid messing around with CMake policy


CMakeLists.txt
```diff
@@ -434,7 +436,8 @@ target_link_libraries(
     fmt::fmt 
     nlohmann_json::nlohmann_json 
     ${STD_FS_LIB} # from helpers.cmake
-    Minuit2::Minuit2
+    Minuit2::Minuit2 #TODO! can we make this private?
```
kferjaoui: as long as `FitModel.hpp` is a public header with Minuit2 member types, I guess the dependency must be PUBLIC

kferjaoui (approved):

PR #304, opened 2026-04-15
allow passing mask to clustervector

### LINKED ISSUE #294: Numpy array back to ClusterVector
I am currently working in python and transforming a ClusterVector from a .clust file into an array, doing some slicing and then converting it back to a ClusterVector. At the moment I am doing it like:

```
with ClusterFile(file) as f:
            for clusters in f:
                arr = np.array(clusters, copy=False)
                ...
                good_clusters = arr[good_mask]
                cv = ClusterVector()
                for i in range(good_clusters.shape[0]):
                    c = Cluster(int(good_clusters['x'][i]), int(good_clusters['y'][i]), good_clusters['data'][i].ravel())
                    cv.push_back(c)
```

It would be nice if it was possible in the future, to only write 

```
cv = ClusterVector(good_clusters)
```

or something similar, to get rid of the for loop.

Thanks!

AliceMazzoleni99:
- passing mask to ClusterVector 
- creates a copy of the ClusterVector


erikfrojdh (approved):
Looks good to me.

PR #305, opened 2026-04-16
Dev/enable custom etas

AliceMazzoleni99:
- Allowing the users more flexibility to play around with custom eta functions without touching the c++ code

- passing vector of eta values to ``transform_eta_values`` 

```
from aare import Interpolator, ClusterVector, Etai, Cluster
import numpy as np 

def custom_eta(cluster_pixel_coordinate_x, cluster_pixel_coordinate_y, cluster_data):
    # dummy custom eta function that just returns the sum of the cluster data
    eta = Etai()
    eta.x = 0.1 # dummy x value
    eta.y = 0.1 # dummy y value
    eta.sum = np.sum(cluster_data) # sum of the cluster data as the "energy
    return eta

# Create a dummy eta distribution and bins
eta_distribution = np.zeros((10, 10, 1)) # dummy eta distribution
etax_bins = np.linspace(0, 1.0, 11)
etay_bins = np.linspace(0, 1.0, 11)
e_bins = np.array([0., 10.]) # dummy energy bins

# Create the interpolator
interpolator = Interpolator(eta_distribution, etax_bins, etay_bins, e_bins)

# Create a dummy cluster vector
cluster_vector = ClusterVector()
cluster_vector.push_back(Cluster(10, 5, np.ones(shape=9, dtype = np.int32)))
cluster_vector.push_back(Cluster(20, 10, np.ones(shape=9, dtype = np.int32)))

# Create dummy etas for the clusters
cluster_array = np.array(cluster_vector)
etas = np.array([custom_eta(cluster["x"], cluster["y"], cluster["data"]) for cluster in cluster_array])

# transform eta values to uniform coordinates 
uniform_coordinates = interpolator.transform_eta_values(etas)

# Interpolate to get the photon coordinates e.g. apply interpolation logic 
photon_coordinates_x = cluster_array["x"] + uniform_coordinates["x"] # add to pixel coordinate 
photon_coordinates_y = cluster_array["y"] + uniform_coordinates["y"] # add to pixel coordinate 

```
advantage: full control over interpolation logic, 
downside: inefficient quite some loops in python
- passing pre computed eta values to interpolate function 
```
Interpolator.interpolate(cluster_vector, etas) 
```
downside: less flexibility in interpolation logic. 
downside: People might misuse it instead of using interpolate directly with a pre compiled eta function implemented in c++

### DIFF
--- a/include/aare/Interpolator.hpp
+++ b/include/aare/Interpolator.hpp
@@ -226,15 +226,15 @@ Interpolator::interpolate(const ClusterVector<ClusterType> &clusters) const {
         photon.y = cluster.y;
         photon.energy = static_cast<decltype(photon.energy)>(eta.sum);
 
+        Coordinate2D uniform_coordinates{};
+
         try {
             // check if eta values are within bounds
-            transform_eta_values(eta);
+            uniform_coordinates = transform_eta_values(eta);
         } catch (const std::runtime_error &e) {
             throw std::runtime_error(
                 fmt::format("{} for cluster: {}", e.what(), cluster_index));
         }
-
-        auto uniform_coordinates = transform_eta_values(eta);
 
         if (EtaFunction == &calculate_eta2<typename ClusterType::value_type,
                                            ClusterType::cluster_size_x,
@@ -311,15 +311,14 @@ std::vector<Photon> Interpolator::interpolate(
         photon.y = cluster.y;
         photon.energy = static_cast<decltype(photon.energy)>(eta.sum);
 
+        Coordinate2D uniform_coordinates{};
         try {
             // check if eta values are within bounds
-            transform_eta_values(eta);
+            uniform_coordinates = transform_eta_values(eta);
         } catch (const std::runtime_error &e) {
             throw std::runtime_error(
                 fmt::format("{} for cluster: {}", e.what(), i));
         }
-
-        auto uniform_coordinates = transform_eta_values(eta);
 
         photon.x += uniform_coordinates.x;
         photon.y += uniform_coordinates.y;
--- a/include/aare/Interpolator.hpp
+++ b/include/aare/Interpolator.hpp
@@ -122,17 +122,6 @@ class Interpolator {
     template <typename T>
     Coordinate2D transform_eta_values(const Eta2<T> &eta) const;
 
-    /**
-     * @brief transforms the eta values to uniform coordinates based on the CDF
-     * ieta_x and ieta_y for a vector of eta values
-     * @tparam T type of eta values
-     * @param etas vector of eta values to transform
-     * @return vector of uniform coordinates {x,y}
-     */
-    template <typename T>
-    std::vector<Coordinate2D>
-    transform_eta_values(const std::vector<Eta2<T>> &etas) const;
-
   private:
     /**
      * @brief bilinear interpolation of the transformed eta values
@@ -219,19 +208,6 @@ Coordinate2D Interpolator::transform_eta_values(const Eta2<T> &eta) const {
     // bilinear_interpolation(ix, iy, ie, eta);
 
     return Coordinate2D{m_ietax(ix, iy, ie), m_ietay(ix, iy, ie)};
-}
-
-template <typename T>
-std::vector<Coordinate2D>
-Interpolator::transform_eta_values(const std::vector<Eta2<T>> &etas) const {
-    std::vector<Coordinate2D> uniform_coordinates;
-    uniform_coordinates.reserve(etas.size());
-
-    for (const auto &eta : etas) {
-        uniform_coordinates.push_back(transform_eta_values(eta));
-    }
-
-    return uniform_coordinates;
 }
 
 template <auto EtaFunction, typename ClusterType, typename Enable>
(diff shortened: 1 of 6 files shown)


include/aare/Interpolator.hpp
```diff
@@ -265,4 +285,49 @@ Interpolator::interpolate(const ClusterVector<ClusterType> &clusters) const {
...
+
+        try {
+            // check if eta values are within bounds
+            transform_eta_values(eta);
+        } catch (const std::runtime_error &e) {
+            throw std::runtime_error(
+                fmt::format("{} for cluster: {}", e.what(), i));
+        }
+
+        auto uniform_coordinates = transform_eta_values(eta);
```
thattil: does this need to be called twice? Could you not split it to validate first?

include/aare/Interpolator.hpp
```diff
@@ -265,4 +285,49 @@ Interpolator::interpolate(const ClusterVector<ClusterType> &clusters) const {
...
+        photon.x = cluster.x;
+        photon.y = cluster.y;
+        photon.energy = static_cast<decltype(photon.energy)>(eta.sum);
+
+        try {
+            // check if eta values are within bounds
+            transform_eta_values(eta);
+        } catch (const std::runtime_error &e) {
+            throw std::runtime_error(
+                fmt::format("{} for cluster: {}", e.what(), i));
```
thattil, line 320: do you want to also give the position too for debugging?

python/src/bind_Interpolator.hpp
```diff
@@ -48,6 +51,34 @@ void register_interpolate(py::class_<aare::Interpolator> &interpolator,
         docstring.c_str(), py::arg("cluster_vector"));
 }
 
+template <typename Type, uint8_t ClusterSizeX, uint8_t ClusterSizeY,
+          typename CoordType = uint16_t>
+void register_interpolate_costum_eta(
```
thattil: typo on the name.. custom_eta?

python/src/bind_Interpolator.hpp
```diff
@@ -48,6 +51,34 @@ void register_interpolate(py::class_<aare::Interpolator> &interpolator,
...
+    py::class_<aare::Interpolator> &interpolator) {
+
+    using ClusterType = Cluster<Type, ClusterSizeX, ClusterSizeY, CoordType>;
+
+    interpolator.def(
+        "interpolate",
+        [](aare::Interpolator &self, const ClusterVector<ClusterType> &clusters,
+           const std::vector<Eta2<Type>> &etas) {
+            auto photons = self.interpolate<Type, ClusterSizeX, ClusterSizeY, CoordType>(clusters, etas);
+            auto *ptr = new std::vector<Photon>{photons};
```
thattil: can one not use a smart pointer for this? and is it better to std::move the photons instead of copy? r can you do directly return self.interpolate?

thattil: can you do self.interfpolate(clusers, etas)? without the template arguments?

python/src/bind_Interpolator.hpp
```diff
@@ -48,6 +51,34 @@ void register_interpolate(py::class_<aare::Interpolator> &interpolator,
...
 
+template <typename Type, uint8_t ClusterSizeX, uint8_t ClusterSizeY,
+          typename CoordType = uint16_t>
+void register_interpolate_costum_eta(
+    py::class_<aare::Interpolator> &interpolator) {
+
+    using ClusterType = Cluster<Type, ClusterSizeX, ClusterSizeY, CoordType>;
+
+    interpolator.def(
+        "interpolate",
```
thattil, line 62: do you want to be explicit in naming it 'interpolate_custom_eta' or something like that?

python/tests/test_InterpolationAPI.py
```diff
@@ -0,0 +1,31 @@
...
+    eta1.y = 0.1
+    eta1.sum = 5
+    eta2 = Etai()
+    eta2.x = 0.1
+    eta2.y = 0.9
+    eta2.sum = 6
+    etas = np.array([eta1, eta2]) # dummy etas for the clusters
+
+    photons = interpolator.interpolate(cluster_vector, etas)
+
```
thattil, line 30: maybe a test to test for size mismatch between clusters and etas.

thattil (approved):
Minor changes or questions.

include/aare/Interpolator.hpp
AliceMazzoleni99: which position?

include/aare/Interpolator.hpp
AliceMazzoleni99: Good catch. Moved definition before try block

python/src/bind_Interpolator.hpp
AliceMazzoleni99: It should be clear from the overload. Maybe I should implicitly define interpolate as interpolate_eta2

python/src/bind_Interpolator.hpp
AliceMazzoleni99: Oh good point added std::move. But if we return directly it will cast it to a list which looses cache locality in python. So return_vector will create a numpy array.

include/aare/Interpolator.hpp
thattil: i meant x and y for the position, but never mind.

PR #306, opened 2026-04-23
Feature/cuda clusterfinder

kferjaoui:
**CUDA Cluster Finder**

*Kernel*

- Shared memory tiling with generalized halo loading for arbitrary cluster sizes (3×3, 5×5, 9×9, …)
- Compile-time stencil geometry enables full loop unrolling of the stencil reduction
- Pedestal subtraction fused into the shared memory load stage
- Non-photon pixels update the pedestal in-kernel; no atomics required on this path
- Lock-free atomic global counter for cluster output across blocks
- ~87.5% occupancy on `sm_89` (43 registers/thread, 0 spills)

*Precision*

- Stencil arithmetic and shared memory use float (`COMPUTE_TYPE` alias)
- Pedestal accumulation stays double to preserve variance accuracy
- FP32 throughput is ~64× higher than FP64 on the RTX 4090
- For 3x3 clusters: The stride-18 shared memory access pattern avoids bank conflicts only with 32-bit values

*Host Wrapper & Transfers*

- Per-stream pinned host staging buffers for truly async DMA (no implicit sync on pageable memory)
- N independent `StreamContext` instances, each owning device buffers, a stream handle, and a pedestal array
- `find_clusters_batched()` distributes frames round-robin across streams
- Cluster payload construction deferred out of the kernel hot path
- Rely on squared comparisons to avoid per-pixel `sqrtf()`

*Python Integration*

- Built as a separate `_aare_cuda.so` to avoid nvcc/gcc ABI conflicts
- Re-exported onto `_aare` so user imports remain `from aare import ClusterFinderCUDA`
- `Cluster` and `ClusterVector` marked `py::module_local()` for pybind11 3 compatibility
- Runtime factory selects CPU or GPU backend; raises RuntimeError on CPU-only builds

*Validation*

- 97.5% cluster match vs. CPU reference on ~90k frames of 400×400 Moench data (9×9 clustering)
- ~134× throughput improvement: 106 FPS CPU vs. 14,214 FPS GPU
- Kernel time: 0.040 ms/frame | PCIe + overhead: 0.030 ms/frame


src/Makefile
lrlunin, line 1: Should not be included in git, I guess

CMakeLists.txt
```diff
@@ -88,6 +93,46 @@ if(AARE_SYSTEM_LIBRARIES)
...
+    set(CMAKE_CUDA_ARCHITECTURES ${AARE_CUDA_ARCHITECTURES})
+  endif()
+
+  message(
+    STATUS "AARE: CUDA support ENABLED (archs=${CMAKE_CUDA_ARCHITECTURES}, "
+           "toolkit=${CUDAToolkit_VERSION})")
+else()
+  message(STATUS "AARE: CUDA support DISABLED (CPU-only build)")
+endif()
+
```
lrlunin, line 115: Duplication of code from lines 96:115
```suggestion

```

src/Makefile
kferjaoui: Yes will be deleted (only for quick tests while developing)

CMakeLists.txt
kferjaoui: Good catch!

.github/workflows/build_conda.yml
```diff
@@ -40,6 +40,9 @@ jobs:
       - name: Disable upload
         run: conda config --set anaconda_upload no
 
+      - name: Install conda-build in base
+        run: conda install -n base conda-build -y
+
```
lrlunin, line 45:
```suggestion
```
fixed in #309

.github/workflows/build_conda_and_deploy_on_main.yml
```diff
@@ -35,6 +35,9 @@ jobs:
       - name: Enable upload
         run: conda config --set anaconda_upload yes
 
+      - name: Install conda-build in base
+        run: conda install -n base conda-build -y
+
```
lrlunin, line 40:
```suggestion
```
fixed in #309

lrlunin (commented):
The issue with failing conda build was resolved in #309

include/aare/clusterfinder_kernel.cuh
```diff
@@ -0,0 +1,355 @@
...
+/// against a host-computed baseline: f32 and f64 agree to 3e-7 in cluster
+/// count, and f32 cuts the 9x9 kernel time by ~40%.
+using COMPUTE_TYPE = float;
+/// Device pedestal storage and running variance update. See COMPUTE_TYPE.
+using DEVICE_PED_TYPE = float;
+
+template <typename ClusterType = Cluster<int32_t, 3, 3>,
+          typename FRAME_TYPE = uint16_t,
+          typename = std::enable_if_t<no_2x2_cluster<ClusterType>::value>>
+__global__ void find_clusters_in_single_frame(
```
tomascelko: As I see it, the whole kernel is about 300 lines, do you think it would be possible to split it into a few functions ? Reasonable compiler would inline them usually, and maybe it would be easier to read.

include/aare/clusterfinder_kernel.cuh
```diff
@@ -0,0 +1,355 @@
...
+    // Squared threshold constants; avoids sqrt at runtime
+    // c2^2 is for the 2x2 quadrant test -- commented out along with Test 2
+    //     itself (see below), whose disabled block is its only reader. Left in
+    //     place rather than deleted so re-enabling Test 2 is one uncomment.
+    // c3^2 is for the full-cluster total test
+    // constexpr int pow2_c2 = ((CSY + 1) / 2) * ((CSX + 1) / 2);
+    constexpr int pow2_c3 = CSX * CSY;
+
+    // Thread/pixel mapping
+    auto col_global =
```
tomascelko: Nice, thanks to this you get coalesced reads :)

include/aare/clusterfinder_kernel.cuh
```diff
@@ -0,0 +1,355 @@
...
+
+    // Cooperative zero-fill
+    for (int idx = static_cast<int>(local_tid); idx < tile_size;
+         idx += static_cast<int>(blockDim.x * blockDim.y)) {
+        shmem[idx] = COMPUTE_TYPE{0};
+    }
+    __syncthreads();
+
+    // OOB flag
+    bool valid_pixel = col_global < static_cast<ssize_t>(ncols) &&
```
tomascelko: Perhaps it makes sense to get rid of ssize_t in geenral (windows doesnt like it usually) in favor of more portable and self documenting types (int64_t, uint32_t...) ? And wouldnt 32 bits suffice for number of columns ?

include/aare/clusterfinder_kernel.cuh
```diff
@@ -0,0 +1,355 @@
...
+    auto shmem_tid =
+        (static_cast<int>(threadIdx.y) + row_radius) * shmem_stride +
+        (static_cast<int>(threadIdx.x) + col_radius);
+
+    // Cooperative zero-fill
+    for (int idx = static_cast<int>(local_tid); idx < tile_size;
+         idx += static_cast<int>(blockDim.x * blockDim.y)) {
+        shmem[idx] = COMPUTE_TYPE{0};
+    }
+    __syncthreads();
```
tomascelko: Nice, this is mandatory

include/aare/clusterfinder_kernel.cuh
```diff
@@ -0,0 +1,355 @@
...
+                       row_global < static_cast<ssize_t>(nrows);
+
+    // ======================================================
+    // Load pedestal-subtracted frame data into shared memory (MIXED PRECISION)
+    // ======================================================
+
+    // Helper: read (frame - pedestal_mean) from global memory, or 0 if OOB.
+    // gr, gc are the global row/col of the pixel to load.
+    // Returns the pedestal-subtracted value.
+    auto load_pixel = [&] __device__(ssize_t gr, ssize_t gc) -> COMPUTE_TYPE {
```
tomascelko: While lambdas are great, why wouldnt we dedicate a proper function to do the job ? One needs to verify if potentital argument passing would make any meaningful slowdown. I would guess it is comaprable to lambda. I admit this might be a matter of preference, feel free to ignore If you like

include/aare/clusterfinder_kernel.cuh
```diff
@@ -0,0 +1,355 @@
...
+        auto gid = gc + ncols * gr;
+        return static_cast<COMPUTE_TYPE>(d_frame[gid]) - d_pd_mean[gid];
+    };
+
+    // A. Interior: every valid thread loads its own pixel
+    if (valid_pixel) {
+        shmem[shmem_tid] = load_pixel(row_global, col_global);
+    }
+
+    // B. Halo regions (Boundaries)
```
tomascelko: This whole section is very repetetive. There must be a way to wrap the for loop into a function without losing any measurable performance. I understand the rationale behind the "branch first" attitude but at least it should be possible to wrap into a function and call it repeatedly.   Hm, for loading data into shared memory and bypassing registers, maybe `cuda::memcpy_async` could be used (?). I know it exists and could probably be used for this purpose, but I haven't tested it myself yet.

include/aare/clusterfinder_kernel.cuh
```diff
@@ -0,0 +1,355 @@
...
+    // float — unlike the raw E[X^2] - E[X]^2 form, where X ~ 4600 makes both
+    // terms ~2e7. NOTE: Keep thresholds squared to avoid one sqrtf() per pixel.
+    DEVICE_PED_TYPE mean_px = d_pd_mean[global_tid];
+    DEVICE_PED_TYPE resid = mean_px - d_pd_off[global_tid]; // E[Y] ~ O(1)
+    DEVICE_PED_TYPE var_px =
+        d_pd_sum2[global_tid] / static_cast<DEVICE_PED_TYPE>(n_pd_samples) -
+        resid * resid;
+    COMPUTE_TYPE rms_sq = static_cast<COMPUTE_TYPE>(
+        var_px > DEVICE_PED_TYPE{0} ? var_px : DEVICE_PED_TYPE{0});
+    COMPUTE_TYPE nSig_sq_rms_sq = m_nSigma * m_nSigma * rms_sq;
```
tomascelko: just out of curiosity, what is the casing rule you are using ? For instance for "nSig_sq_rms_sq". Moreover, COMPUTE_TYPE seems to evaluate to float, and I think you are not assigning to this variable anywhere, I'd declare it as const.This can indicate to compiler that this is just an alias to computed result. I think there are multiple such variables in this function that could be declared as const. (Maybe the compiler can already guess it, but it wouldnt hurt for sure)

include/aare/clusterfinder_kernel.cuh
```diff
@@ -0,0 +1,355 @@
...
+    //  3. Total significance:         total > c3 * nSigma * rms
+    //     -> distributed events where the full cluster sum is significant
+
+    bool is_photon = false;
+
+    // Test 1: single-pixel significance
+    //     max_val > nSigma * rms
+    // is equivalent to:
+    //     max_val > 0 && max_val^2 > nSigma^2 * rms^2
+    if (max_val > 0.0f && max_val * max_val > nSig_sq_rms_sq) {
```
tomascelko: Is comparing against float constants safe ? Maybe some small epsilon is safer to compare against (0.00001 or so)?

include/aare/clusterfinder_kernel.cuh
```diff
@@ -0,0 +1,355 @@
...
+    COMPUTE_TYPE max_val = -HUGE_VAL; // double inf; narrows to float if needed
+
+    // // Quandrants
+    // PEDESTAL_TYPE tl = PEDESTAL_TYPE{0};   // top-left quadrant  (ir<=0,
+    // ic<=0) PEDESTAL_TYPE tr = PEDESTAL_TYPE{0};   // top-right quadrant
+    // (ir<=0, ic>=0) PEDESTAL_TYPE bl = PEDESTAL_TYPE{0};   // bottom-left
+    // (ir>=0, ic<=0) PEDESTAL_TYPE br = PEDESTAL_TYPE{0};   // bottom-right
+    // (ir>=0, ic>=0)
+
+#pragma unroll
```
tomascelko: Now, this is an interesting one. You said you have 6blocks/SM with 42 registers.
Q1: For which compute capability is it ? Recently I noticed some of my kernels use diferent number of registers depending on how old compute capability I was targeting (older compute capabilities are worse at unrolling sometimes).
 6 blocks - this is enough to saturate 4070, but not enough to saturate A100 (can do up to 8blocks for max theoretical occupancy).
Since I dont think this kernel requires 42 x 4B of variables, probably less, I would guess that additional variables come from unrolling of the loop (pre-fetching)
Q2: What happens if you remove the pragma unroll, does compiler decide to unroll it by itself ? (how does it affect performance and register count?) If you target also A100 or similar GPUs, could it make sense to only partially unroll `#pragma unroll n` and maybe save some registers in favor of higher occupancy? Just a tip.

include/aare/clusterfinder_kernel.cuh
```diff
@@ -0,0 +1,355 @@
...
+    for (int ir = -row_radius; ir <= row_radius; ++ir) {
+#pragma unroll
+        for (int ic = -col_radius; ic <= col_radius; ++ic) {
+            COMPUTE_TYPE val = shmem[shmem_tid + ir * shmem_stride + ic];
```
tomascelko: For example here I think also this can maybe be const ?

tomascelko:
Overall, I think It's a nice and reasonably optimized code. So far, I only got through the kernel and added a couple of comments. Some of them I consider a matter of preference; feel free to ignore the ones which you don't agree with. I hope to return again at some point to see the rest of the implementation. Keep up the good work, and don't hesitate to correct me if I misunderstood anything :)

tomascelko:
Actually I have a question. How hard would it be to implement API to cuda cluster finder, one that would be able to do the following:
```
template<typename ValueType, typename ClusterType, typename PedestalType>
class CudaClusterFinderApi
{
    void processFrames(PedestalType* pedestal_matrix, const ValueType* frame_values, ClusterType* out_clusters, cudaStream_t* stream) {/*do the processing of batch of frames on a given stream, update the pedestal matrix and return the clusters*/}
}
```
I'm asking because if this API exists, I could include it into the sparspix-sls project I'm working on as another algorithm option. The memory & stream management would be handled outside of the API, I would just need the isolated clustering & pedestal updating logic. Just a thought, feel free to say no :D

include/aare/clusterfinder_kernel.cuh
kferjaoui: `ssize_t` is used across the aare (and its core) so only changing the kernel won't win portability. But, for performance sake, changing to `int32_t` may be a reasonable call. 
Also we still need a signed type as the halo loads rely on `row_global - i >= 0` checks

include/aare/clusterfinder_kernel.cuh
kferjaoui: I dont think `memcpy_async` would help here as the `load_pixel` performs a fused load/substraction operation, not a copy

include/aare/clusterfinder_kernel.cuh
kferjaoui: Q1. Yes it is 8.9 for Ada arch (RTX4090). And your analysis is correct.
Q2. Good point: Partial unrolling, using `#pragma unroll 1` on the outer loop for e.g. , does improve theoretical and measured occupancies for bigger clusters. Nice call!

P.S: Higher occupancy and better performance are, however, not always the same thing ;)

include/aare/clusterfinder_kernel.cuh
kferjaoui: In principle, you are right and more than half are comments (which I agree bloat the header). But (even for me who wrote it) I find the description useful to put the intent into words (at least for now). Plus, the algorithm itself may still evolve...
During the merge, I can improve readability in case others complain

include/aare/clusterfinder_kernel.cuh
kferjaoui: The 2 cases you mention (`max_val > 0.0f` and `total > 0.0f`) are sign guards so I'd rather not introduce an epsilon

kferjaoui:
> Actually I have a question. How hard would it be to implement API to cuda cluster finder, one that would be able to do the following:
> 
> ```
> template<typename ValueType, typename ClusterType, typename PedestalType>
> class CudaClusterFinderApi
> {
>     void processFrames(PedestalType* pedestal_matrix, const ValueType* frame_values, ClusterType* out_clusters, cudaStream_t* stream) {/*do the processing of batch of frames on a given stream, update the pedestal matrix and return the clusters*/}
> }
> ```
> 
> I'm asking because if this API exists, I could include it into the sparspix-sls project I'm working on as another algorithm option. The memory & stream management would be handled outside of the API, I would just need the isolated clustering & pedestal updating logic. Just a thought, feel free to say no :D

I think it is a very good point; separating device mem management from the ClusterFinder's API is the way to expose a general interface that would work independently of the underlying algorithm. 

I am attempting a code refactoring in a seperate branch for testing atm. Will share when ready.

PR #317, opened 2026-05-27
PixelHistogram

erikfrojdh:
Multi threaded filling of per pixel histograms for example for detector calibration

1. PixelHistogram - Generic variant expects already pedestal subtracted data
2. PedestalTrackingHistogram - Terrible name, useful class. Keeps it's own pedestal and does conversion and pedestal tracking in the worker threads. 

The logic when to update the pedestal is taken from the current Jungfrau calibration and needs to match it at least for the time being. 

**Still needs**
- [ ] Tests
- [ ] Documentation


include/aare/PixelHistogramImpl.hpp
```diff
@@ -0,0 +1,124 @@
...
+    for (int row = 0; row < m_rows; ++row) {
+        for (int col = 0; col < m_cols; ++col) {
+            fill(row, col, frame(row, col));
+        }
+    }
+}
+
+template <typename T, typename StorageType>
+void PixelHistogramImpl<T, StorageType>::fill(int row, int col, T value) {
+    // TODO! add out of bounds check on row and col???
```
thattil: please add this check?

include/aare/hist/PixelHistogramImpl.hpp
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+}
+
+template <typename T, typename StorageType>
+void PixelHistogramImpl<T, StorageType>::fill(int row, int col, T value) {
+    // TODO! add out of bounds check on row and col???
+
+    if (value < m_xmin || value >= m_xmax) {
+        return;
```
thattil, line 102: should this silently return? Shoudl it not raise or log or counted separately?

erikfrojdh: This is quite common that you try to fill outside the range of the axis. Other histogram classes typically have a overflow and underflow bin that does effectively the same. So I would say this is expected behaviour.

include/aare/PixelHistogramImpl.hpp
erikfrojdh: It's in now. And I have a method called fill_unchecked that can be used internally (when we fill from NDView) and outside for performance critical stuff

include/aare/hist/PixelHistogram.hpp
```diff
@@ -0,0 +1,89 @@
...
+    const AxisType xmax_;
+    std::vector<Hist> partial_hists_;
+    // Cumulative row offsets so that thread t owns rows
+    //     [row_offsets_[t], row_offsets_[t + 1])
+    // Length is n_threads_ + 1; partition is balanced (the first
+    // rows_ % n_threads_ threads get one extra row each).
+    std::vector<int> row_offsets_;
+
+    // Thread pool members
+    std::vector<std::thread> workers_;
```
thattil, line 41: Maybe something for the future (not for this PR) would be to extract the workers or coordination logic out of here.

include/aare/hist/PixelHistogramImpl.hpp
```diff
@@ -0,0 +1,136 @@
...
+    }
+    int bin = static_cast<int>((value - m_xmin) * m_scale);
+    // Guard against floating-point rounding pushing val just below
+    // xmax to bin == n_bins.
+    if (bin >= m_n_bins) {
+        bin = m_n_bins - 1;
+    }
+    ++m_values(row, col, bin);
```
thattil, line 116: Can this overflow? Or is it handled in NDArray?

include/aare/hist/PixelHistogram.hpp
```diff
@@ -0,0 +1,89 @@
...
+    std::mutex work_mutex_;
+    std::condition_variable work_cv_;
+    std::condition_variable done_cv_;
+    const NDView<AxisType, 2> *current_image_;
+    std::atomic<int> completed_threads_;
```
thattil: Do you need atomic if you are always using mutex?

thattil (commented):
Have only gone through Impl() and PixelHistogram. LGTM, minor comments/questions. Need to go through the others.

include/aare/hist/PixelHistogram.hpp
erikfrojdh: Into some generic multi threading class?

include/aare/hist/PixelHistogramImpl.hpp
erikfrojdh: Yes, it could overflow. Added checks and benchmarked that they did not impact performance. values now stuck at max.

include/aare/hist/PixelHistogram.hpp
erikfrojdh: You're correct changed them to normal ints.

include/aare/hist/PixelHistogram.hpp
thattil: Yes, it might be useful to reuse this. Anyways, not for this PR.

include/aare/hist/PedestalTrackingPixelHistogram.hpp
```diff
@@ -0,0 +1,133 @@
...
+    using AxisType = float; // TODO: template on pedestal type if needed
+    using FrameType = uint16_t;
+
+  private:
+    using Hist = PixelHistogramImpl<AxisType, StorageType>;
+    using AsyncQueue = ProducerConsumerQueue<NDArray<FrameType, 2>>;
+
+    // What kind of fan-out work the worker pool should currently do.
+    // Set under work_mutex_; read by worker_loop after wakeup.
+    enum class WorkKind { PushPedestal, UpdateMean, FillWithThreshold };
```
thattil, line 33: 👍

include/aare/hist/PedestalTrackingPixelHistogram.hpp
```diff
@@ -0,0 +1,133 @@
...
+    WorkKind current_work_kind_;
+    const NDView<FrameType, 2> *current_image_;
+    const std::vector<NDView<FrameType, 2>> *current_images_;
+    int completed_threads_;
+    std::atomic<bool> stop_workers_;
+    int work_generation_;
+
+    // Serialises all fan-outs: `push_pedestal_no_update`, `update_mean`,
+    // and the async coordinator's batch fill dispatches.
+    // Always the outermost lock; work_mutex_ is taken briefly inside it.
```
thattil, line 65: 👍

include/aare/hist/PixelHistogram.hpp
```diff
@@ -0,0 +1,344 @@
...
+    void coordinator_loop();
+    // Fan a single image out to the worker pool and block until every
+    // worker has merged its row band. Only ever called by the coordinator
+    // thread, so no caller-serialisation lock is needed.
+    void dispatch(const NDView<AxisType, 2> &image);
+    int row_start(int thread_id) const;
+    int row_count(int thread_id) const;
+
+  public:
+    PixelHistogram(int rows, int cols, int n_bins, double xmin, double xmax,
```
thattil: Shouldn't the xmin and xmax be of AxisType instead of double as in PedestalTrackingPixelHistogram?

src/hist/PedestalTrackingPixelHistogram.cpp
```diff
@@ -0,0 +1,584 @@
...
+    : rows_(rows), cols_(cols), n_threads_(n_threads), xmin_(xmin), xmax_(xmax),
+      current_work_kind_(WorkKind::FillWithThreshold), current_image_(nullptr),
+      current_images_(nullptr), completed_threads_(0), stop_workers_(false),
+      work_generation_(0), max_batch_size_(max_pending / 3), n_sigma_(n_sigma) {
+    if (rows_ < 1 || cols_ < 1 || n_bins < 1) {
+        throw std::invalid_argument("PedestalTrackingPixelHistogram requires "
+                                    "positive rows, cols and bins");
+    }
+    if (n_threads < 1) {
+        throw std::invalid_argument(
```
thattil, line 27: There is some duplication between PedestalTrackingPixelHistorgram and PixelHistorgram constructors (validation, dimension, hist parmaeter initialization). Maybe an option would be to move some of these  into a something like HistogramConfig (rows, cols, bins, xmax, xmin) and HistogramLayout (row offsets, nthreads). Then the constructor would look cleaner.  With the destructors, as mentioned before all the  worker logic could be encapsulated in resusable helper. Now any bug found here will have to be fixed in both the classes. Just my opinion and doesnt have to be in this PR.

include/aare/hist/PixelHistogram.hpp
```diff
@@ -0,0 +1,344 @@
...
+    std::atomic<bool> coordinator_busy_{false};
+    std::chrono::microseconds async_wait_{100};
+
+    // Private worker thread method
+    void worker_loop(int thread_id);
+    void coordinator_loop();
+    // Fan a single image out to the worker pool and block until every
+    // worker has merged its row band. Only ever called by the coordinator
+    // thread, so no caller-serialisation lock is needed.
+    void dispatch(const NDView<AxisType, 2> &image);
```
thattil, line 66: Minor: Same naming consistency as PedestalTrackingPIxelHistogram. Its a private function with an underscore at the end. Just for this function.

src/hist/PedestalTrackingPixelHistogram.cpp
```diff
@@ -0,0 +1,584 @@
...
+        completed_threads_ = 0;
+        current_work_kind_ = kind;
+        current_image_ = image;
+        current_images_ = nullptr;
+        ++work_generation_;
+    }
+    work_cv_.notify_all();
+    {
+        std::unique_lock<std::mutex> lock(work_mutex_);
+        done_cv_.wait(lock,
```
thattil, line 128: If a worker throws, maybe we should catch this and increment completed_threads to avoid a deadlock?

include/aare/Pedestal.hpp
```diff
@@ -110,6 +112,27 @@ template <typename SUM_TYPE = double> class Pedestal {
         }
     }
 
+    template <typename T>
+    void push_with_threshold(const NDView<T, 2> frame,
```
thattil, line 116: should this be a const ref?

include/aare/Pedestal.hpp
```diff
@@ -110,6 +112,27 @@ template <typename SUM_TYPE = double> class Pedestal {
...
+                             const NDView<SUM_TYPE, 2> threshold) {
+        assert(frame.size() == m_rows * m_cols);
+
+        // TODO! move away from m_rows, m_cols
+        if (frame.shape() != std::array<ssize_t, 2>{m_rows, m_cols}) {
+            throw std::runtime_error(
+                "Frame shape does not match pedestal shape");
+        }
```
thattil, line 124: do we need to validate threshold shape as well?

python/src/bind_PedestalTrackingPixelHistogram.hpp
```diff
@@ -0,0 +1,245 @@
...
+                 n_sigma: Sigma multiplier used as the gate for the
+                     pedestal-update side effect of
+                     fill_async(): a pixel sample is
+                     pushed back into the pedestal estimate iff
+                     ``abs(residual) < n_sigma * cached_std``. Set to
+                     ``0.0`` to disable the pedestal update and get
+                     histogram-only async behaviour (default: 1.0).
+                     Also exposed live via the ``n_sigma`` property.
+             )",
+            py::arg("rows"), py::arg("cols"), py::arg("n_bins"),
```
thattil: do you want to make this also kw_only, just as in PixelHistogram binding?

thattil (approved):

include/aare/hist/PixelHistogram.hpp
erikfrojdh: Good catch, updated it.

include/aare/Pedestal.hpp
erikfrojdh: Not needed since NDView is a lightweight type. Basically just pointer and shape. Another option would be to pass an NDAarry as const ref.

python/src/bind_PedestalTrackingPixelHistogram.hpp
erikfrojdh: Makes sense!

include/aare/Pedestal.hpp
erikfrojdh: Yes, added it.

src/hist/PedestalTrackingPixelHistogram.cpp
erikfrojdh: I need to think about how to do it properly. But right now I'm not sure anything would throw in the worker loop...

src/hist/PedestalTrackingPixelHistogram.cpp
erikfrojdh: This should be high on the list once people start using. For now I think it's good to give people something to play with.

include/aare/hist/PixelHistogram.hpp
erikfrojdh: Removed the _

PR #319, opened 2026-06-03
Fix version

thattil:
- install VERSION file and copy to build
- fix lookup for VERSION file in case of error
- update documentation for installing python bindings


erikfrojdh (approved):
Looks good to me. Doesn't anymore rely on the fact that we do an in source build.

PR #320, opened 2026-06-09
Conditional use of Minuit2

erikfrojdh:
If AARE_FETCH_MINUIT is set to OFF we first look for a standalone Minuit2 and if that is not found we try to find Minuit2 as a part of ROOT. 

In both cases we make an alias to allow for simpler use of the target later. 

It still doesn't solve the issue that we install Minuit to when we fetch it but that can be addressed in a separate PR. 

closes #316

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,6 +150,8 @@ else()
 endif()
 
 if(AARE_FETCH_MINUIT2)
+  # We are building Minuit2 from sources.
+
   # Patch minuit2 to avoid messing with cmake policies
   set(MINUIT2_PATCH_COMMAND git apply
                             ${CMAKE_CURRENT_SOURCE_DIR}/patches/minuit2.patch)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -176,8 +178,21 @@ if(AARE_FETCH_MINUIT2)
 
   set_property(TARGET Minuit2Math PROPERTY POSITION_INDEPENDENT_CODE ON)
   set_property(TARGET Minuit2 PROPERTY POSITION_INDEPENDENT_CODE ON)
+
+  add_library(aare::Minuit2 ALIAS Minuit2)
```
AliceMazzoleni99, line 182: is the actual target called Minuit2::Minuit2?

AliceMazzoleni99 (approved):
small comment. Havent tested it with root pre-installed.

CMakeLists.txt
erikfrojdh: Both actually =). I had a look at the source code of Minuit and they use an alias
```
# Add alias for direct inclusion with add_subdirectory
add_library(Minuit2::Minuit2 ALIAS Minuit2)
```

In total Minuit exposes these targets under the Minuit2 namespace
```
Minuit2Common Minuit2Math Minuit2
```

But for set_property we need to use the "real" target and not the alias.

PR #322, opened 2026-06-10
Bugfix: NDArary/NDView overflow in calculating number of elements

erikfrojdh:
Fixed number of elements calculation that caused integer overflow. Now return size_t instead of int.


thattil (approved):
LGTM

PR #324, opened 2026-06-11
Dev/matterhorn decoder

AliceMazzoleni99:
- reshape image directly in decoder such that first dimension is num counters 
- take into account chip artefact in decoder.


erikfrojdh (approved):
Single counter images with (1,256,256) could trip up some people but is consistent with multi counters.

PR #328, opened 2026-06-12
len() and pedestal subtraction

erikfrojdh:
Collection of small improvements in usability: 

- NDView works for large arrays
- Direct subtraction of Pedestal from np.array
- len() support for python bindings of files

### DIFF
--- a/python/src/pedestal.hpp
+++ b/python/src/pedestal.hpp
@@ -13,9 +13,9 @@ namespace py = pybind11;
 
 template <typename SUM_TYPE>
 void define_pedestal_bindings(py::module &m, const std::string &name) {
-    
+
     py::class_<Pedestal<SUM_TYPE>>(m, name.c_str(), py::buffer_protocol())
-	.def(py::init<int, int, int>())
+        .def(py::init<int, int, int>())
         .def(py::init<int, int>())
         .def("mean",
              [](Pedestal<SUM_TYPE> &self) {
--- a/python/src/pedestal.hpp
+++ b/python/src/pedestal.hpp
@@ -13,10 +13,9 @@ namespace py = pybind11;
 
 template <typename SUM_TYPE>
 void define_pedestal_bindings(py::module &m, const std::string &name) {
-    auto pedestal =
-        py::class_<Pedestal<SUM_TYPE>>(m, name.c_str(), py::buffer_protocol());
-
-    pedestal.def(py::init<int, int, int>())
+    
+    py::class_<Pedestal<SUM_TYPE>>(m, name.c_str(), py::buffer_protocol())
+	.def(py::init<int, int, int>())
         .def(py::init<int, int>())
         .def("mean",
              [](Pedestal<SUM_TYPE> &self) {
--- a/python/src/pedestal.hpp
+++ b/python/src/pedestal.hpp
@@ -5,6 +5,7 @@
 
 #include <cstdint>
 #include <filesystem>
+#include <pybind11/numpy.h>
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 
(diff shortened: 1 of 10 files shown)


src/CtbRawFile.cpp
```diff
@@ -57,6 +57,8 @@ size_t CtbRawFile::image_size_in_bytes() const {
 
 size_t CtbRawFile::frames_in_file() const { return m_master.frames_in_file(); }
 
+size_t CtbRawFile::total_frames() const { return m_master.frames_in_file(); }
+
```
AliceMazzoleni99, line 61: whats total_frames() in comoarison to frames_in_file(). Can you document the use cases. I find frames_in_file a better name.

python/src/pedestal.hpp
```diff
@@ -5,15 +5,18 @@
...
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 
 namespace py = pybind11;
 
 template <typename SUM_TYPE>
 void define_pedestal_bindings(py::module &m, const std::string &name) {
-    py::class_<Pedestal<SUM_TYPE>>(m, name.c_str())
-        .def(py::init<int, int, int>())
+    auto pedestal =
```
AliceMazzoleni99: what is the advantage of this syntax over py::calss?  I get a lot of warnings for all the lambda functions

python/src/pedestal.hpp
```diff
@@ -50,6 +53,23 @@ void define_pedestal_bindings(py::module &m, const std::string &name) {
                  *std = self.std();
                  return return_image_data(std);
              })
+        .def(
+            "__array_ufunc__",
+            [](py::object self, py::object ufunc, const std::string &method,
+               py::args inputs, py::kwargs kwargs) -> py::object {
+                if (method != "__call__" || inputs.size() != 2 ||
+                    inputs[1].ptr() != self.ptr() ||
+                    py::cast<std::string>(ufunc.attr("__name__")) !=
```
AliceMazzoleni99, line 61: Minor: shouldnt we then just have .def("__sub__") if any other operations are not supported anyway?

AliceMazzoleni99: `__sub__`

AliceMazzoleni99 (commented):
Added some minor comments. Or things I havent understood. Looks good otherwise. Nice to directly allow subtraction without converting to numpy array.

src/CtbRawFile.cpp
erikfrojdh: ` virtual size_t total_frames() const = 0;` is what we currently have in the FileInterface that we use for the generic interface. Which of course can be changed since not too many people use aare. 

In our master file it's unfortunately more complicated:

- `Total Frames`  - Total number of frames and triggers expected (i.e. what the user sets)
-  `Frames in File` - Number of frames actually written to file by Receiver 0 (can be more or less...)

For the API I think total_frames should mean the amount of frames in the file. But we can discuss changing the names in another PR do make it more clear. Maybe even in the master file...

Will make an issue.

python/src/pedestal.hpp
erikfrojdh:
```python
frame = np.zeros((512,1024), dtype = np.uint16)
pd = Pedestal(512,1024.....)


data = frame-pd #we want to support having pd on the right side so have to trigger from the numpy array over the ufunc
```

In the code we check that there is a subtraction
```cpp
py::cast<std::string>(ufunc.attr("__name__")) !=
                        "subtract")

```

Maybe it's possible to implement over rsub instead but not sure it brings an improvement.

python/src/pedestal.hpp
erikfrojdh: The warning is because of shadowing of the name. Didn't see it since clang is not as picky as gcc. 
But here it's not needed to name pedestal so we can just use py::class.. and then directly chained .def 

Named it first since I wanted to reuse the name but then didn't...

Reverted syntax

src/CtbRawFile.cpp
AliceMazzoleni99: Oke but then shouldnt it ve m_master.total_frames_expected();

erikfrojdh: No that's the point, in the API we use`total_frames` for the number of frames in the file. I agree that maybe we should change it, but I would not change it in this PR. It's been in the public interface on File, RawFile and JungfrauDataFile since the start of aare. 

Here I only added it to the CtbRawFile for consistency.

AliceMazzoleni99: Ah oke got it. The File class uses total_frames.

AliceMazzoleni99 (approved):

